PR #6, opened 2016-03-09
Fix comparison between signed and unsigned integer warning

ddemidov:


jrprice:
Thanks for the fix - sorry for the delay.

PR #9, opened 2016-04-28
Rework make files to build on Mac OS X

ghulands:
OpenCL on OS X is still at 1.2 support. These changes wrap the build for 2.x in If checks.


jrprice:
I'm uncomfortable with these changes. I personally use a Macbook as my primary machine and have never had any issues with building and testing the 2.x versions of these C++ bindings on OS X. You don't need an implementation of OpenCL 2.x to build or test this header, just the OpenCL 2.x headers and a library to link against, both of which are now also open source.

Although Apple haven't released OpenCL 2.x support, third-party implementations of OpenCL that run on OS X are out there, some of which are starting to add OpenCL 2.0 support (e.g. pocl). I don't think we should be disabling 2.x on OS X just because Apple themselves aren't supporting it.

jrprice:
Closing - as mentioned above we do not want to disable 2.x on macOS.

PR #10, opened 2016-04-28
Have dependencies in submodules

ghulands:
Makes building out of the box quicker.


jrprice:
As discussed in #9, we don't want to accept the proposed changes for macOS. I agree with the spirit of adding submodules for CMock and Unity though, since this makes it simpler to get this project built. I've pushed a commit which does this in a manner consistent with some of the other Khronos projects  (e.g. using `external` as the directory name for external dependencies).

PR #11, opened 2016-04-29
Added 8 missing Get<> 1.2 specializations  + 1 return type fix

Teybeo:
I'm putting these fixes out but I would like to address a bigger problem:

As I was making these fixes, I noticed that these missing enums added in CL 1.2 were not present in the OpenCL 1.2 C++ Wrapper spec, which might explain why there are missing from the cl.hpp headers.

The thing is: the spec is not technically incorrect because it only includes the "values that differ in return type between the OpenCL C API and the OpenCL C++ API" and only refers to the OpenCL C spec for the ones that don't need special C++ types like string and vectors instead of C pointers.

As all the 8 missing enums had primitive return types, I guess that what happened is that the person/s who updated the cl.hpp missed that subtlety and only added in the header the values that were added in the CL 1.2 C++ spec tables, and thus missed these 8 ones.

Anyways, I don't think it can merged as-is because:
- I didn't updated the cl2.hpp which might require special care since some of those might have been deprecated in 2.0, 2.1 or 2.2.
-  I didn't run the tests due to non-installed dependencies (even though the changes are trivial).

### DIFF
--- a/input_cl.hpp
+++ b/input_cl.hpp
@@ -1421,7 +1421,6 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
 #if defined(CL_VERSION_1_1)
 #define __PARAM_NAME_INFO_1_1(F) \
     F(cl_context_info, CL_CONTEXT_NUM_DEVICES, cl_uint)\
-    \
     F(cl_device_info, CL_DEVICE_PREFERRED_VECTOR_WIDTH_HALF, cl_uint) \
     F(cl_device_info, CL_DEVICE_NATIVE_VECTOR_WIDTH_CHAR, cl_uint) \
     F(cl_device_info, CL_DEVICE_NATIVE_VECTOR_WIDTH_SHORT, cl_uint) \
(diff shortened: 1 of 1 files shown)


input_cl.hpp
```diff
@@ -1421,6 +1421,7 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
 #if defined(CL_VERSION_1_1)
 #define __PARAM_NAME_INFO_1_1(F) \
     F(cl_context_info, CL_CONTEXT_NUM_DEVICES, cl_uint)\
+    \
```
jrprice: This looks like an accidental (unnecessary change)?

input_cl.hpp
```diff
@@ -1458,16 +1462,21 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
...
+    F(cl_device_info, CL_DEVICE_LINKER_AVAILABLE, cl_bool) \
+    F(cl_device_info, CL_DEVICE_BUILT_IN_KERNELS, STRING_CLASS)\
+    F(cl_device_info, CL_DEVICE_PRINTF_BUFFER_SIZE, ::size_t) \
+    F(cl_device_info, CL_DEVICE_PREFERRED_INTEROP_USER_SYNC, cl_bool) \
     F(cl_device_info, CL_DEVICE_PARENT_DEVICE, cl_device_id) \
+    F(cl_device_info, CL_DEVICE_PARTITION_MAX_SUB_DEVICES, cl_uint) \
     F(cl_device_info, CL_DEVICE_PARTITION_PROPERTIES, VECTOR_CLASS<cl_device_partition_property>) \
+    F(cl_device_info, CL_DEVICE_PARTITION_AFFINITY_DOMAIN, cl_device_affinity_domain) \
     F(cl_device_info, CL_DEVICE_PARTITION_TYPE, VECTOR_CLASS<cl_device_partition_property>)  \
     F(cl_device_info, CL_DEVICE_REFERENCE_COUNT, cl_uint) \
```
jrprice: With the way you've rearranged these lines, there's now a trailing `\` which causes compilation failures for me.

Teybeo: Fixed

input_cl.hpp
Teybeo: Fixed

jrprice:
Thanks for the fixes, and apologies again that this has taken so long.

Teybeo:
Better late than never :slightly_smiling_face:

Issue #8, opened 2016-04-11
Defect: includes cl::Kernel::getInfo() string contains an explicit NUL byte


trbauer:
For information where device or kernel info of type `char[]` are returned, 
e.g.  `clGetKernelInfo(...,CL_KERNEL_FUNCTION_NAME,&size)`
  (tested on NVidia and Intel SDK's, not sure about AMD)
The size value returned _includes_ the terminating byte for these SDKs. Consequently, when the `cl.hpp` helper function copies out  the string data it ends up including the explicit NUL byte as part of the string (and it's size).

``` C++
  // Specialized GetInfoHelper for STRING_CLASS params
  template <typename Func>
  inline cl_int getInfoHelper(Func f, cl_uint name, STRING_CLASS* param, long) {
    ...
    if (param) {
        param->assign(value.begin(), value.end()); // <<< COPIES NUL byte explicitly into std::string
    }
}
```

This leads to a fairly difficult debugging task since the following expression evaluates to `false`.
   `(k.getInfo<CL_KERNEL_FUNCTION_NAME>() == std::string("foo"))`
And worse, some debuggers typically don't show the explicit null byte in the `std::string`.

The OpenCL spec is not terribly clear on whether the NUL byte should be included for (char[]) info types. I confirmed this behavior on Intel and NVidia's SDKs at least, but I am not sure what AMD and other do and hence I don't know what the "right" solution is, but I think the current approach adds a "sharp edge" to this otherwise wonderful C++ wrapper API.

coleb:
We just ran into this problem as well when upgrading to the latest version of cl.hpp. This terminating NUL byte really breaks code like this: 

```
std::string uname;
uname += platform.getInfo<CL_PLATFORM_VERSION>();
uname += " ";
uname += device.getInfo<CL_DEVICE_NAME>();
```

The following diff fixes our test cases and gets the rest of our system up and running again with the new header: 

```
diff --git a/input_cl.hpp b/input_cl.hpp
index ad62b0f..7a5e134 100644
--- a/input_cl.hpp
+++ b/input_cl.hpp
@@ -1243,6 +1243,9 @@ inline cl_int getInfoHelper(Func f, cl_uint name, STRING_CLASS* param, long)
     if (param) {
         param->assign(value.begin(), value.end());
     }
+    if (!param->empty() && param->back() == '\0') {
+      param->resize(param->size() - 1);
+    }
 #endif
     return CL_SUCCESS;
 }
[...]

PR #12, opened 2016-05-02
Issue 8

coleb:
Fixes the std::string bugs reported in issue #8. 

Also adds rudimentary ability to build test suites on OSX. Though looks like I may have just duplicated that effort with @ghulands. Oh well. 

Wish there was travis integration testing for this stuff. Only tested on my OSX machine.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,12 @@ if( CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID STREQUAL "Clang" )
   set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -pthread")
 endif()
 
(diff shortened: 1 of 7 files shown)


jrprice:
I was able to reproduce the bugs described in issue #8 with `cl.hpp`, but *not* with `cl2.hpp`. This leads me to believe these bugs were fixed when we last encountered NULL terminator issues in this function (fixed in a few commits around July/August last year). If this is the case, I would prefer to backport the same fixes that were made in `cl2.hpp` for consistency sake, rather than re-fixing this with a different approach.

If there *is* still an issue with the very latest `cl2.hpp`, can you provide an example that reproduces the problem?

As for OS X support, please see [my comment on a previous pull request](https://github.com/KhronosGroup/OpenCL-CLHPP/pull/9#issuecomment-222489714).

jrprice:
Closing since this was fixed via #39 instead.

PR #13, opened 2016-05-30
Add cmake build options for docs, examples & tests

9prady9:
When this repository is included as external cmake project in another projects build process, having an option to disable building docs/examples/tests would be nice. The PR adds these options.


jrprice:
Seems perfectly reasonable to me, thanks.

PR #14, opened 2016-06-01
Add install target for CL headers

9prady9:
This may not be absolutely needed change but having the headers at specified location rather than at cmake-generated path is more useful for down stream projects.

Default `CMAKE_INSTALL_PREFIX` is `/usr/include` on linux, thus resulting in appropriate installation of  `CL/*` files to `/usr/include` path.


9prady9:
@jrprice I have also added a check for function `getVersion`, please check.

jrprice:
Looks good.

Sorry for taking a while to get round to this!

9prady9:
That's fine, thank you for taking care of it :+1:

PR #15, opened 2016-06-13
use "${CMAKE_COMMAND} -E copy" instead of cp

petrkalos:
Generating cl2.hpp on Windows is failing because "cp" command is missing, thus this merge request is replacing cp with "${CMAKE_COMMAND} -E copy"


jrprice:
Looks good, thanks!

Issue #18, opened 2016-08-03
Creating 2D image from buffer in OpenCL 1.2 implementations


bsochack:
Creating 2D images from buffer should be allowed when cl_khr_image2d_from_buffer is supported by a device.

Currently the image2d constructor is guarded by this #ifdef:

#if CL_HPP_TARGET_OPENCL_VERSION >= 200

PR #19, opened 2016-08-03
creating image 2D from buffer should be also allowed if cl_khr_image2…

bsochack:
…d_from_buffer extension is supported

Fix for issue: https://github.com/KhronosGroup/OpenCL-CLHPP/issues/18


jrprice:
This change looks good to me.

Could you also add an entry for this macro to the `Optional extension support` section in the comments at the top of the header (line 47), so that people know that the macro exists without having to dig through the whole header file?

bsochack:
Done, please check now

jrprice:
Great, thanks!

PR #21, opened 2016-08-13
Create LICENSE.txt

bkmgit:
Copied text from source files to make the material easier to find.


jrprice:
That's the correct license, but please remove the C-style comment characters as is typical for license files and then I can accept it.

bkmgit:
Removed the C-style comment characters.

PR #22, opened 2016-10-03
add const qualifier to enqueueMarkerWithWaitList, enqueueBarrierWithW…

petrkalos:
Following the pattern of the rest of the header I can't see why these 3 methods shall be non-const, thus I am proposing to add the const qualifier.


jrprice:
Thanks!

PR #28, opened 2017-03-08
Fixing CommandQueue constructor for improved OpenCL < 2.0 compatibility

rtobar:
A runtime problem occurs when an application is compiled with `CL_HPP_TARGET_OPENCL_VERSION >= 200` and `CL_HPP_MINIMUM_OPENCL_VERSION < 200`. At compile time the CommandQueue constructor happily runs because it was using `clCreateCommandQueueWithProperties`, but when the code was run against a platform supporting OpenCL < 2.0 then the same function, which was only introduced in OpenCL 2.0, was being called instead of `clCreateCommandQueue` (usually causing a segfault).

This fix is inspired by the constructor of the Image* classes, which use a combination of macros and  runtime version checking to choose which function should be called.

Any feedback would be appreciated. I'm also not rushing and adding myself to the `\author` doxy tag, but if this gets eventually merged it would be much appreciated.

### DIFF
--- a/input_cl2.hpp
+++ b/input_cl2.hpp
@@ -6699,31 +6699,49 @@ public:
         }
         else {
             Device device = context.getInfo<CL_CONTEXT_DEVICES>()[0];
+            bool useWithProperties;
+
+#if CL_HPP_TARGET_OPENCL_VERSION >= 200 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
+            // Run-time decision based on the actual platform
+            {
+                cl_uint version = detail::getContextPlatformVersion(context());
+                useWithProperties = (version >= 0x20000); // OpenCL 2.0 or above
+            }
+#elif CL_HPP_TARGET_OPENCL_VERSION >= 200
+            useWithProperties = true;
+#else
+            useWithProperties = false;
+#endif
 
 #if CL_HPP_TARGET_OPENCL_VERSION >= 200
-            cl_queue_properties queue_properties[] = {
-                CL_QUEUE_PROPERTIES, properties, 0 };
-            if ((properties & CL_QUEUE_ON_DEVICE) == 0) {
-                object_ = ::clCreateCommandQueueWithProperties(
-                    context(), device(), queue_properties, &error);
-            }
-            else {
-                error = CL_INVALID_QUEUE_PROPERTIES;
-            }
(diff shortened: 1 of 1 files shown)


rtobar:
Any news on whether this can be merged back?

input_cl2.hpp
```diff
@@ -6699,31 +6699,49 @@ class CommandQueue : public detail::Wrapper<cl_command_queue>
...
+                detail::errHandler(error, __CREATE_COMMAND_QUEUE_WITH_PROPERTIES_ERR);
+                if (err != NULL) {
+                    *err = error;
+                }
             }
-#else
-            object_ = ::clCreateCommandQueue(
-                context(), device(), properties, &error);
+#endif // CL_HPP_TARGET_OPENCL_VERSION >= 200
+#if CL_HPP_MINIMUM_OPENCL_VERSION < 200
```
jrprice: I think this can just be an `#else`. If the above `CL_HPP_TARGET_OPENCL_VERSION >= 200` condition fails, then the minimum will *always* be `< 200`. In this case, you also don't need the `if (!useProperties)` check below.

Same with all the other places where you've made this change.

rtobar: Hi James,

Thanks for taking the time to review this and thank you for your comments.

I think the second ``#if`` you highlighted cannot be turned into an ``#else`` (and therefore the ``if (!useProperties)`` cannot be removed either). You are right that when pointing out that ``CL_HPP_TARGET_OPENCL_VERSION < 200`` implies ``CL_HPP_MINIMUM_OPENCL_VERSION < 200``, but it's not necessarily true the other way around (and is the situation I found myself in when I found this problem, I have minimum set to 1.1, but target is 2.0); therefore they are not mutually exclusive and should be accounted for separately. That's also why the ``useProperties`` is defined in all cases and used inside the final ``#if CL_HPP_MINIMUM_OPENCL_VERSION < 200``.

I know the code is can be a bit hard to read this way. Still, if you have a look at the Image2D class constructors you will find a similar logic applied there, which is what I based my solution on.

Please let me know what do you think, and if I should give the patch another try.

jrprice:
OK, since this matches the `Image2D` approach I'm fine with this. Sorry once again for the delay!

rtobar:
Thanks a lot!

PR #29, opened 2017-03-24
Small const reference fixes on range-based for loops

Milania1:
Just a small change where I replaced the types of the for loops with const references.


jrprice:
Merged - sorry for the wait!

PR #31, opened 2017-09-06
Fix redundant Wrapper::get()

chfast:
When returning cl_type by value do not add additional const modifier. This fixes GCC warning "type qualifiers ignored on function return type".

### DIFF
--- a/input_cl2.hpp
+++ b/input_cl2.hpp
@@ -1699,10 +1699,7 @@ public:
 
     cl_type& operator ()() { return object_; }
 
-    const cl_type get() const { return object_; }
-
-    cl_type get() { return object_; }
-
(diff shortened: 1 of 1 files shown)


jrprice:
I notice that in your [`ethminer` commit](https://github.com/ethereum-mining/ethminer/commit/3b44ecc501f99eff7da2f959c79104b8e4db19aa) you made this change in the `Wrapper<cl_device_id>` specialization (around line 1830), whereas this PR is proposing the change for the main `Wrapper` class (around line 1702). Was this a mistake? Presumably we want this change in both places anyway.

chfast:
Probably a mistake. I will fix it soon.

jrprice:
Fix merged in #34.

PR #34, opened 2017-11-30
Fixed version of PR31

sethtroisi:
@chfast Never showed back up in https://github.com/KhronosGroup/OpenCL-CLHPP/pull/31

This modifies only the cl_device_id version of wrapper.

### DIFF
--- a/input_cl2.hpp
+++ b/input_cl2.hpp
@@ -1828,9 +1828,7 @@ public:
 
     cl_type& operator ()() { return object_; }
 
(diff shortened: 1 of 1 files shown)


jrprice:
Sorry for the delay, merged - thanks for the fix!

sethtroisi:
Happy that it got merged!
Thanks for review.

chfast:
Can you make a release with this fix?

PR #39, opened 2018-02-15
Fix NULL Terminator in string class

### LINKED ISSUE #8: Defect: includes cl::Kernel::getInfo() string contains an explicit NUL byte

Flamefire:
Fixes #8
Backport of 8e9aebb (No NULL byte) and c43a7ed (Return empty string)

Similar to PR #12 but fixes only the bug. Uses a backport of the cl2.hpp header fix suggested by @jrprice 

CCing @coleb

### DIFF
--- a/input_cl.hpp
+++ b/input_cl.hpp
@@ -1233,15 +1233,20 @@ inline cl_int getInfoHelper(Func f, cl_uint name, STRING_CLASS* param, long)
         return err;
     }
 
-    // std::string has a constant data member
-    // a char vector does not
-    VECTOR_CLASS<char> value(required);
-    err = f(name, required, value.data(), NULL);
-    if (err != CL_SUCCESS) {
-        return err;
+    if (required > 0) {
+        // std::string has a constant data member
+        // a char vector does not
+        VECTOR_CLASS<char> value(required);
+        err = f(name, required, value.data(), NULL);
+        if (err != CL_SUCCESS) {
+            return err;
+        }
+        if (param) {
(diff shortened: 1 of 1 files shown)


jrprice:
Merged - thanks!

PR #47, opened 2018-06-18
Fix catch of non-polymorphic cl::Error, fix uninitialized error value

TimTronic:
Newest version of gcc warned about the non-polymorphic catch. I also noticed that error is never set if the exception is caught (because the throw happens before the assign).


jrprice:
Sorry for the delay, finally had a chance to test this and it looks fine, thanks for the contribution!

Issue #48, opened 2018-07-04
cl2.hpp/cl.hpp  does not work for OpenCL 1.1


gallickgunner:
So according to the `cl2.hpp`

> C++ bindings for OpenCL 1.0 (rev 48), OpenCL 1.1 (rev 33),  OpenCL 1.2 (rev 15) and OpenCL 2.0 (rev 29)

However when I tried targeting my project for openCL 1.1 using these defines
```
#define CL_MINIMUM_OPENCL_VERSION 110
#define CL_HPP_MINIMUM_OPENCL_VERSION 110
#define CL_HPP_TARGET_OPENCL_VERSION 110
#define CL_TARGET_OPENCL_VERSION 110
```
It throws a bunch of errors shown below. The errors remain the same if i use `cl.hpp`

![errors](https://user-images.githubusercontent.com/40718428/42260767-1915f2fa-7f7f-11e8-9e15-26f8d22e95b9.png)

The project compiles fine if I change the last two defines to `1.2.0` or greater. I have tried both, using the precompiled release version which gives `cl2.hpp` and compiling the `cl2.hpp` myself using cmake.

Any help would be appreciated.

Additional Information.
OS: Win 7 64-bit
IDE: Codeblocks
Compiler : Mingw-w64 gcc 8.1.0 (32- bit version)  Have also tried (mingw32 gcc 6.3.0)
GPU: Radeon hd 7950 (OpenCL 2.0 supported)

jrprice:
Fixed now that #49 is merged.

PR #49, opened 2018-07-07
Fix Issues when targeting OpenCL 1.1 or 1.0

gallickgunner:
The commit fixes 2 issues.
1) When targeting **OpenCL 1.0** the compiler throws an error for `enqueueReadBufferRect` saying it hasn't been declared in scope. This is because `enqueueXBufferRect` calls have been introduced from  
 **OpenCL 1.1**. The commit fixes this by adding a version check for these calls.

2) When targeting **OpenCL 1.1**  the compiler throws an error regarding `CL_DEVICE_DOUBLE_FP_CONFIG`. For detailed description of this issue, see: https://github.com/KhronosGroup/OpenCL-CLHPP/issues/48

This issue arises due to missing ` #defines` in OpenCL-C headers namely `cl.h` and `cl_ext.h`. Briefly describing, the macro `CL_DEVICE_DOUBLE_FP_CONFIG` is missing from `cl_ext.h` and only defined for **OpenCL 1.2 or greater** in `cl.h`.
  
Whereas both `CL_DEVICE_DOUBLE_FP_CONFIG` and `CL_DEVICE_HALF_FP_CONFIG` should be available at all OpenCL versions. 

This issue #48 is fixed by the following pending pull request to "OpenCL-Headers" registry

https://github.com/KhronosGroup/OpenCL-Headers/pull/28

However both `cl.hpp` and `cl2.hpp` define this parameter name from **OpenCL 1.1** onwards. This means it is still not available for OpenCL 1.0 when using the C++ bindings header. 

The commit fixes this 2nd issue by removing both parameter names `DOUBLE/HALF_FP_CONFIG`  from `PARAM_NAME_INFO_1_1` and adding them under `PARAM_NAME_INFO_1_0`

### DIFF
--- a/input_cl.hpp
+++ b/input_cl.hpp
@@ -1337,6 +1337,8 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     F(cl_device_info, CL_DEVICE_MEM_BASE_ADDR_ALIGN, cl_uint) \
     F(cl_device_info, CL_DEVICE_MIN_DATA_TYPE_ALIGN_SIZE, cl_uint) \
     F(cl_device_info, CL_DEVICE_SINGLE_FP_CONFIG, cl_device_fp_config) \
+    F(cl_device_info, CL_DEVICE_DOUBLE_FP_CONFIG, cl_device_fp_config) \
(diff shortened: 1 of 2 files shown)


jrprice:
Looks good, thanks for the contribution!

Issue #25, opened 2016-12-09
Build cl.hpp for releases, not just cl2.hpp


oddhack:
Could the release builds be extended to include generating the old cl.hpp as well? That would let me remove it from the OpenCL registry, where it's the only header file still stored there instead of in github.

jrprice:
Personally I'd like to get rid of the old `cl.hpp` completely, given that the new `cl2.hpp` supports all the 1.x versions of OpenCL as well and has many other general improvements. That would mean that the registry could just point users to `cl2.hpp` for all versions of OpenCL.

I'll see if I can get agreement to do this from the working group.

oddhack:
That would be even better. I didn't realize cl2.hpp was usable for CL 1.x (largely because the name suggests otherwise :-) )

jrprice:
I agree, the name may be misleading here. It's not backwards compatible with the previous version which is why there was a name change when adding the 2.0 support. We may want to consider another name change if we want to present this as a 'one size fits all' header.

kpet:
I think a name change is indeed the way to go. The current state is pretty confusing and it wouldn't be hard to keep backwards compatibility with applications using `cl2.hpp`. `opencl.hpp`? Any other suggestions?

rajhlinux:
I was getting compiler errors and never knew that cl2.hpp is the replacement and updated header file for cl.hpp

So many github projects like "Caffe" (Deep learning framework) and many others have not updated to cl2.hpp and getting this error on 2022... At least this thread clear things up.

bashbaug:
Would it be helpful to describe this more explicitly in the README?  It currently describes cl2.hpp (and how it has been renamed to opencl.hpp), but there is currently no description for cl.hpp.

It would be nice to add more clarification to the Doxygen docs also, though it is at least mentioned [there](http://khronosgroup.github.io/OpenCL-CLHPP/):

> There are numerous compatibility, portability and memory management fixes in the new header as well as additional OpenCL 2.0 features. As a result the header is not directly backward compatible and for this reason we release it as [opencl.hpp](https://github.khronos.org/OpenCL-CLHPP/opencl_8hpp.html) rather than a new version of cl.hpp.

PR #54, opened 2018-12-11
Add notices for cl.hpp deprecation

jrprice:
The OpenCL WG have previously agreed that we wish to deprecate the older `cl.hpp` header in favour of `cl2.hpp`, which also supports OpenCL 1.x.

This relates to #25. The one outstanding niggle here is that the name of the newer header is a little misleading, and suggests that it only supports OpenCL 2.x.

### DIFF
--- a/README.txt
+++ b/README.txt
@@ -6,9 +6,10 @@ Doxgen documentation for the cl2.hpp header is available here:
 
 
 Components:
-  input_cl.hpp:
+  input_cl.hpp (DEPRECATED):
     Acts as the master source for the 1.x version of the header.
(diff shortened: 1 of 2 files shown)


bashbaug:
These changes look good to me.  I like the notice in cl.hpp directing users to the new header file.

I'm not thrilled with the name "cl2.hpp", since the "new" header does work with non-2.x versions of OpenCL, but I can't think of a better name.

jrprice:
Agreed to accept on OpenCL WG call 12th Feb 2019.

We'll remain open to proposals for a better name for `cl2.hpp`.

PR #55, opened 2018-12-12
Handle CL_TARGET_OPENCL_VERSION more rigorously

jrprice:
Produce a warning if `CL_TARGET_OPENCL_VERSION` is already defined and is lower than the requested `CL_HPP_TARGET_OPENCL_VERSION`.

If `CL_TARGET_OPENCL_VERSION` is already defined and is *greater* than `CL_HPP_TARGET_OPENCL_VERSION`, you might end up with some deprecated function warnings in the C++ header that are not expected, for the same reason. Should we actually warn if there is any mismatch between these two macros at all?

### DIFF
--- a/input_cl2.hpp
+++ b/input_cl2.hpp
@@ -447,8 +447,17 @@
 # undef CL_HPP_TARGET_OPENCL_VERSION
 # define CL_HPP_TARGET_OPENCL_VERSION 200
 #endif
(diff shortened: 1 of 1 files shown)


jrprice:
Scratch my comment there - `cl2.hpp` will define the `CL_USE_DEPRECATED_*_APIS` macros if necessary, so shouldn't be an issue.

PR #56, opened 2019-02-07
input_cl: fix GCC warning

martingkelly:
Currently with -Wall, GCC is throwing the following warning:

warning: catching polymorphic type ‘class cl::Error’ by value
[-Wcatch-value=]. Fix this by catching by reference instead.


jrprice:
Thanks for the fix!

Please note that `cl.hpp` has now been deprecated in favour of `cl2.hpp`, which also supports OpenCL 1.x and has many other improvements (including already having fixed this bug, for example).

Issue #62, opened 2019-06-19
Add continuous integration


kpet:
- [x] Run the header generation script
- [x] Build the tests
- [x] Run the tests
- [x] Anything else?

kpet:
We now have basic continuous integration up and running. I'm going to close this. I've created #78 to try and capture a detailed list of the configurations that we want to test.

PR #63, opened 2019-07-19
Basic Travis CI config

kpet:
Not running the tests for now as some of them are failing.

Contributes to #62.


kpet:
Thanks for the review! I tend to use `$TOP` for conciseness but, as you say, it doesn't really matter. Merging.

PR #67, opened 2019-09-16
Fix failing tests and enable testing in Travis

kpet:
Since we build with CL_HPP_MINIMUM_OPENCL_VERSION defined to 100,
the bindings are querying the platform to decide whether to use
clCreateCommandQueue or clCreateCommandQueueWithProperties. We
thus need to make the stub function for clGetPlatformInfo report
2.0 when building with CL_HPP_TARGET_OPENCL_VERSION >= 200 to
mirror the queue creation function check.

Also allow calls to clCreateContextFromType with no properties on
Apple platforms to mirror the logic in the bindings.

Contributes to #62.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,3 +24,4 @@ script:
           -DOPENCL_INCLUDE_DIR=${TOP}/OpenCL-Headers
           ..
   - make -j2
+  - ctest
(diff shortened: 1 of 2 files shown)


bashbaug:
LGTM - merging.

PR #68, opened 2019-09-16
Make it possible to use the repository as a submodule directly

kpet:
The new cl2.hpp is not generated so move it as is to include/CL.

The old cl.hpp was the only header that was truly generated.
Commit the generated version and add automation to ensure that
the committed generated version matches the output of the
generation script with the committed template file.


bashbaug:
This looks great, thank you for doing this work.

Rebased and pushed to master as c5063c34675155fffc20185e74ba82a9dfd5dc84.

bashbaug:
One minor comment: I think we need to update the README after these changes.  I've opened #71 so we don't lose track of this.

kpet:
Yes, I'd spotted that. I thought I'd fix on rebase. I'll submit a PR.

PR #70, opened 2019-09-20
ULT fix: need to return a 2.0 platform to use 2.0 queue creation function

bashbaug:
This change fixes a failing cl2hpp ULT.  When the ULT returned 1.2 as its platform version, the headers were using the OpenCL 1.x clCreateCommandQueue API rather than the stubbed clCreateCommandQueueWithProperties API.  Fix is to return 2.0 as the platform version when the ULT is testing the OpenCL 2.x clCreateCommandQueueWithProperties API.


kpet:
This is a duplicate of https://github.com/KhronosGroup/OpenCL-CLHPP/pull/67

bashbaug:
Whoops, sorry about that.  Closing this PR and reviewing #67 instead.

PR #72, opened 2019-09-21
OpenCL 2.2 Support

### LINKED ISSUE #30: OpenCL 2.2 support?

bashbaug:
Adds OpenCL 2.2 support to the C++ bindings.  Specifically:

Adds support for the new queries:

* CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT
* CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT

Also adds support for the new APIs:

* clSetProgramReleaseCallback
* clSetProgramSpecializationConstant

Tested ULTs on Windows and Linux.

Fixes #30.

### DIFF
--- a/include/CL/cl2.hpp
+++ b/include/CL/cl2.hpp
@@ -6795,23 +6795,6 @@ public:
             __SET_PROGRAM_SPECIALIZATION_CONSTANT_ERR);
     }
 
-    /*! \brief Sets a boolean SPIR-V specialization constant.
-     *
-     *  Wraps clSetProgramSpecializationConstant().
-     */
-    template <>
-    cl_int setSpecializationConstant(cl_uint index, const bool &value)
-    {
-        cl_uchar ucValue = value ? CL_UCHAR_MAX : 0;
-        return detail::errHandler(
-            ::clSetProgramSpecializationConstant(
-                object_,
-                index,
-                sizeof(ucValue),
-                &ucValue),
-            __SET_PROGRAM_SPECIALIZATION_CONSTANT_ERR);
-    }
-
     /*! \brief Sets a SPIR-V specialization constant.
      *
      *  Wraps clSetProgramSpecializationConstant().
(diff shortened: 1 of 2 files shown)


include/CL/cl2.hpp
```diff
@@ -6729,6 +6754,62 @@ class Program : public detail::Wrapper<cl_program>
...
+                user_data),
+            __SET_PROGRAM_RELEASE_CALLBACK_ERR);
+    }
+
+    /*! \brief Sets a SPIR-V specialization constant.
+     *
+     *  Wraps clSetProgramSpecializationConstant().
+     */
+    template <typename T>
+    typename std::enable_if<!std::is_pointer<T>::value, cl_int>::type
```
kpet: Why did you disable it for pointers?

include/CL/cl2.hpp
```diff
@@ -6850,6 +6931,22 @@ inline vector<vector<unsigned char>> cl::Program::getInfo<CL_PROGRAM_BINARIES>(c
     return binariesVectors;
 }
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 220
+// Template specialization for clSetProgramSpecializationConstant
+template <>
+inline cl_int cl::Program::setSpecializationConstant(cl_uint index, const bool &value)
+{
+    cl_uchar ucValue = value ? CL_UCHAR_MAX : 0;
```
kpet: This is a bit of an arbitrary choice. Have you considered making this configurable? (We do have many configuration options already though...)

include/CL/cl2.hpp
bashbaug: The SPIR-V spec for [OpSpecConstant](https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpSpecConstant) says:

> Result Type must be a scalar integer type or floating-point type.

I take this to mean that a spec constant cannot be a pointer, so pointers should be disabled.  Does that sound correct?

include/CL/cl2.hpp
bashbaug: The OpenCL spec says for [clSetProgramSpecializationConstant](https://www.khronos.org/registry/OpenCL/specs/2.2/html/OpenCL_API.html#clSetProgramSpecializationConstant):

> If a specialization constant is a boolean constant, spec_value should be a pointer to a cl_uchar value. A value of zero will set the specialization constant to false; any other value will set it to true.

I take this to mean that the actual value being set is fairly arbitrary, so long as it is nonzero.  I don't have a strong opinion which value to pick.

include/CL/cl2.hpp
kpet: Given that the OpenCL entrypoint just allows raw data to be copied (and doesn't really care about type), one might argue that declaring an integer on the SPIR-V side and storing, say an SVM pointer that could be cast back to a pointer after, is technically valid. That being said, people doing weird things like this can still use the other overload and I think I prefer what you've done.

include/CL/cl2.hpp
kpet: Thinking about this a bit more, I'm wondering whether we shouldn't align on Vulkan and use a 32-bit type for the host/device interface. See https://renderdoc.org/vkspec_chunked/chap9.html#pipelines-specialization-constants.

bashbaug: I believe that this implementation with a `uchar` is consistent with the current OpenCL API spec.  I'd be OK changing this to a 32-bit type instead, but I think we'd need to update the OpenCL API spec as well.

kpet: I hadn't spotted the mention of the size/values in the API spec and your first reply. No reason to change anything then.

PR #74, opened 2019-09-23
Don't paralellise Travis builds

kpet:
We seem to be missing dependencies in the build system.

Signed-off-by: Kévin Petit <kevin.petit@arm.com>


kpet:
This is trivial, merging.

PR #77, opened 2019-10-23
Fix a few build warnings

### LINKED ISSUE #45: const cl_type get() type qualifiers warning
When compiling with this parameters `-Wall -Wextra -std=c++11 -O2 -Os -s -DNDEBUG`, I got 

> cl2.hpp:1829:5: warning: type qualifiers ignored on function return type [-Wignored-qualifiers]
>      const cl_type get() const { return object_; }

kpet:
- Remove a couple of unused variables
- Remove ignored qualifier on a return type

Fixes #45

### DIFF
--- a/include/CL/cl2.hpp
+++ b/include/CL/cl2.hpp
@@ -1805,9 +1805,6 @@ public:
 
     cl_type get() const { return object_; }
 
-    cl_type get() { return object_; }
-
-
 protected:
     template<typename Func, typename U>
     friend inline cl_int getInfoHelper(Func, cl_uint, U*, int, typename U::cl_type);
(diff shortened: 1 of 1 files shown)


include/CL/cl2.hpp
```diff
@@ -1803,7 +1803,7 @@ class Wrapper
 
     cl_type& operator ()() { return object_; }
 
-    const cl_type get() const { return object_; }
+    cl_type get() const { return object_; }
```
bashbaug: I think this is OK - since the wrapper class wraps OpenCL handles it's always OK to return a copy of the handle by value even when the wrapper object is "const".  After this change though, do we need both this function and the non-const overload below, or can the non-const overload be removed?

kpet: Good catch, I can't think of a good reason to keep the non-const overload, I'll remove it.

jrprice:
LGTM 👍

bashbaug:
Great, thanks!  Merging.

PR #79, opened 2019-11-06
Add support for device queries from two Arm extensions

kpet:
Add support for the info queries from the following extensions:

cl_arm_core_id
cl_arm_job_slot_selection


bashbaug:
LGTM.  I'll wait another day or so just in case there are other comments, then merge.

PR #81, opened 2020-01-14
Fix the cmake command in README.md

ejricha:
The cmake command was missing the path to the source directory (..)


ejricha:
I believe I have signed the CLA now, please let me know if there is anything else I need to sign. I assume it's just waiting for the Travis CI builds and tests to pass before it pulls.

kpet:
Yes, you did. Just waiting for Travis CI builds now.

PR #82, opened 2020-02-11
C++ bindings for cl_khr_extended_versioning

bashbaug:


include/CL/cl2.hpp
```diff
@@ -1454,6 +1464,10 @@ CL_HPP_PARAM_NAME_INFO_1_2_DEPRECATED_IN_2_0_(CL_HPP_DECLARE_PARAM_TRAITS_)
 CL_HPP_PARAM_NAME_DEVICE_FISSION_(CL_HPP_DECLARE_PARAM_TRAITS_);
 #endif // CL_HPP_USE_CL_DEVICE_FISSION
 
+#if defined(CL_NAME_VERSION_MAX_NAME_SIZE_KHR)  // cl_khr_extended_versioning
```
kpet, line 1467: I'd prefer if we tested for `cl_khr_extended_versioning` but I realise I forgot to add it. I've opened https://github.com/KhronosGroup/OpenCL-Headers/pull/85 to fix this.

bashbaug: I switched to the `cl_khr_extended_versioning` define after merging https://github.com/KhronosGroup/OpenCL-Headers/pull/85 - thanks!

tests/test_cl2hpp.cpp
```diff
@@ -470,13 +470,7 @@ void testCreatePipe()
...
 }
 
-#if defined(_WIN32)
-#define CL_API_CALL     __stdcall
-#else
-#define CL_API_CALL
-#endif
-
-static cl_int CL_API_CALL clGetKernelSubGroupInfo_testSubGroups(cl_kernel kernel,
+static cl_int clGetKernelSubGroupInfo_testSubGroups(cl_kernel kernel,
```
kpet, line 473: This looks like a rebase gone wrong.

include/CL/cl2.hpp
kpet: Thanks!

tests/test_cl2hpp.cpp
```diff
@@ -616,12 +611,26 @@ static void CL_CALLBACK test_program_release_callback(
 {
 }
 
+static cl_int clSetProgramReleaseCallback_set(
```
kpet, line 614: I'm guessing you didn't intend these changes to be part of the PR.

bashbaug: No, this was intended.  I can move this change into a separate PR if desired, but this change is needed to fix the tests on 32-bit Windows due to the way `CL_CALLBACK` is defined and handled by these unit tests.

kpet: It would have been better as a separate PR indeed but, given it's fairly small/straightforward, I'm fine with accepting the PR as is.

PR #85, opened 2020-03-12
Updates cl2.hpp and cl.hpp to handle case of platform with zero devices without error.

Noerr:
Need to more carefully handle the case of a platform detected with zero devices.  In the first call to clGetDeviceIDs, this will return code CL_DEVICE_NOT_FOUND and set n=0 if no devices are reported by the platform. ( https://www.khronos.org/registry/OpenCL/sdk/1.2/docs/man/xhtml/clGetDeviceIDs.html )
The second call to clGetDeviceIDs cannot be allowed occur if n=0, otherwise CL_INVALID_VALUE is reported.
A platform with zero devices is not an error condition.  Device enumeration should simply move on.

### DIFF
--- a/include/CL/cl.hpp
+++ b/include/CL/cl.hpp
@@ -2267,17 +2267,21 @@ public:
             return detail::errHandler(CL_INVALID_ARG_VALUE, __GET_DEVICE_IDS_ERR);
         }
         cl_int err = ::clGetDeviceIDs(object_, type, 0, NULL, &n);
-        if (err != CL_SUCCESS) {
+        if (err != CL_SUCCESS && err != CL_DEVICE_NOT_FOUND) {
             return detail::errHandler(err, __GET_DEVICE_IDS_ERR);
         }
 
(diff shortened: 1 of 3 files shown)


Noerr:
I guess I don't understand what this CI check is doing.  Reviewing the failing script, this draws my attention to input_cl.hpp which has the same issue as my patches to cl2.hpp.
Which one should I change @bashbaug ?

bashbaug:
> Reviewing the failing script, this draws my attention to input_cl.hpp which has the same issue as my patches to (correction) cl.hpp.

Ah yes, I'd forgotten about this change.

In short, `cl.hpp` is a generated file, but it's checked into this repo to ease integration into other projects.  The CI job is testing that output (generated file) is consistent with the input.  So, you should modify the input file `input_cl.hpp`, then commit both the input file and the output file, and the CI job should pass.

Sorry about that, I should have caught this as part of the review.

bashbaug:
Thanks!  Looks like everything is passing now.

PR #90, opened 2020-03-14
Fix test build with optimizations

aaronpuchert:
When building with optimizations on, `test_cl2hpp` segfaulted with both
GCC and Clang. Both compilers tell us why: if a function with non-void
return type does not return via return statement, this is undefined
behavior.

Luckily this can be reliably prevented at compile-time via
`-Werror=return-type`, so we add that flag.


aaronpuchert:
There is some overlap with #82, but I think it's good to keep this separate and prevent such issues at compile-time.

PR #91, opened 2020-03-24
Change license to Apache 2.0

jrprice:
As agreed by the OpenCL WG on the 2020-03-17 teleconference.


bashbaug:
Thanks!  Merging as discussed in the March 24th teleconference.

PR #94, opened 2020-03-30
Detect availability of noexcept via _MSVC_LANG

### LINKED ISSUE #93: Warning C26439 when compiling in Visual Studio 2019
When using `cl2.hpp` in Visual Studio 2019 I'm getting a lot of [C26439](https://docs.microsoft.com/en-us/cpp/code-quality/c26439) IntelliSense Warnings ("This kind of function may not throw. Declare it 'noexcept'.").

The problem seems to be, that at [Line 532](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L532) it is not detected, that `nothrow` is supported, because Visual Studio doesn't set `__cplusplus` to an up to date value by default. (Only if [`/Zc:__cplusplus`](https://docs.microsoft.com/en-us/cpp/build/reference/zc-cplusplus) comandline option is given.)
```
#if (__cplusplus >= 201103L)
#define CL_HPP_NOEXCEPT_ noexcept
#else
#define CL_HPP_NOEXCEPT_
#endif
```

Could Line 532 be changed to this?
```
#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )
```

If `_MSVC_LANG` is not defined it will be evaluated as 0. ([section 16.1 paragraph 4: "After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers and keywords, except for true and false, are replaced with the pp-number 0"](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf#page=403)) If it is defined, than [it's at least Visual Studio 2015 Update 3](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros), which is AFAIK does support `noexcept` (["noexcept operator was introduced and is supported in Visual Studio 2015"](https://github.com/MicrosoftDocs/cpp-docs/blob/8bc0a0e703fd454c42a70965d548708fe503a44b/docs/cpp/exception-specifications-throw-cpp.md)).

__&lt;Edit>__
`noexcept` was added in the [November 2013 CTP (customer technology preview) of the Visual C++ Compiler](https://devblogs.microsoft.com/cppblog/announcing-the-visual-c-compiler-november-2013-ctp/).
__&lt;/Edit>__

This doesn't seem to remove all C26439 warnings but at least more than half of them.

__&lt;Edit2>__
Note: Although exception specifications in Visual C++ are [only completely standard conform if `/std:c++17` is used](https://devblogs.microsoft.com/oldnewthing/20180928-00/?p=99855), it is not necessary to test for `_MSVC_LANG >= 201703L`, because the behavior of `noexcept` is standard conform even without `/std:c++17`.
__&lt;/Edit2>__

TSlivede:
If compiling with Visual Studio, __cplusplus does not hold
an updated value by default, therefore test _MSVC_LANG.

Fixes #93

### DIFF
--- a/include/CL/cl2.hpp
+++ b/include/CL/cl2.hpp
@@ -529,7 +529,7 @@
 #include <CL/opencl.h>
 #endif // !__APPLE__
 
-#if (__cplusplus >= 201103L)
+#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )
 #define CL_HPP_NOEXCEPT_ noexcept
 #else
 #define CL_HPP_NOEXCEPT_


bashbaug:
Hrmph.  I looked through cl2.hpp looking for other places where we are checking for `__cplusplus` and found:

```c
// Need to do both tests here because for some reason __cplusplus is not 
// updated in visual studio
#if (!defined(_MSC_VER) && __cplusplus < 201103L) || (defined(_MSC_VER) && _MSC_VER < 1700)
#error Visual studio 2013 or another C++11-supporting compiler required
#endif
```

In the "old" cl.hpp we also have:

```c
#if (_MSC_VER >= 1700) || (__cplusplus >= 201103L)
```

and:

```c
#if (__cplusplus >= 201103L)
```

I think we should merge this PR as-is, but then go back and handle these checks more uniformly.  @jprice, does this sound reasonable?  Thanks!

TSlivede:
I don't think, that the checks can be made much more uniformly, as Visiual Studio 2013 did not yet have `_MSVC_LANG`, so `_MSC_VER` has to be used. However, the test could be made in the style of `cl.hpp` - so replace
```
#if (!defined(_MSC_VER) && __cplusplus < 201103L) || (defined(_MSC_VER) && _MSC_VER < 1700)
```
by
```
#if !(__cplusplus >= 201103L) || _MSC_VER >= 1700)
```
That would look cleaner and easier to understand in my opinion.

---

The other test, whose change I proposed with this pull request can however not use `_MSC_VER` because just using an appropriately new version of Visual Studio doesn't mean, that an sufficiently new version of the language is selected, so `_MSVC_LANG` has to be used there.

TSlivede:
Another thing, that I just noticed: The mentioned test checks, if Visual Studio is at least version 1700 which corresponds to Visual Studio 2012 according to [this](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros). The error message however says Visual Studio 2013...

jrprice:
I'm not a Visual Studio user, so not sure I've quite understood the issues.

> Visiual Studio 2013 did not yet have `_MSVC_LANG`

So it sounds like *all* of these tests would need to check `_MSC_VER` first. We essentially want:
"if using VS 2015 or newer, then check _MSVC_LANG >= 201103L, else check _cplusplus >= 201103L"
right?

TSlivede:
Oh sorry, I didn't repeat [this explanation](https://github.com/KhronosGroup/OpenCL-CLHPP/issues/93) here:

> If `_MSVC_LANG` is not defined it will be evaluated as 0. ([section 16.1 paragraph 4: "After all replacements due to macro expansion and the defined unary operator have been performed, all remaining identifiers and keywords, except for true and false, are replaced with the pp-number 0"](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2012/n3337.pdf#page=403)) If it is defined, than [it's at least Visual Studio 2015 Update 3](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros), which is AFAIK does support `noexcept` (["noexcept operator was introduced and is supported in Visual Studio 2015"](https://github.com/MicrosoftDocs/cpp-docs/blob/8bc0a0e703fd454c42a70965d548708fe503a44b/docs/cpp/exception-specifications-throw-cpp.md)).

Essentially, we don't need to do such a complicated test because the c and c++ standards specify, that this
```
#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )
```
is essentially equivalent to
```
#if ((defined(__cplusplus)?__cplusplus:0) >= 201103L || (defined(_MSVC_LANG)?_MSVC_LANG:0) >= 201103L )
```
but the former is much easier to read.

jrprice:
Got it, this PR is fine with me then, thanks for clarifying.

Agree with Ben that it'd be nice to be consistent. Given what you've just said, I don't understand your previous comment about why we can't just make all of these tests do `#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )` (or the negation of this).

TSlivede:
The test
```
#if (!defined(_MSC_VER) && __cplusplus < 201103L) || (defined(_MSC_VER) && _MSC_VER < 1700)
```
*can* be changed to the better readable
```
#if !(__cplusplus >= 201103L || _MSC_VER >= 1700)
```
but it can not be changed to an equivalent test that uses `_MSVC_LANG`.

---

However, because of your question, I just thought, that
```
#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )
```
could possibly be changed into a test that uses `_MSC_VER`. That would be more consistent to the other test.

I just have to look up some things, but please __don't merge__ this before I write again.

Thanks for the Idea.

jrprice:
> ```
> #if (!defined(_MSC_VER) && __cplusplus < 201103L) || (defined(_MSC_VER) && _MSC_VER < 1700)
> ```
> [...]
> can not be changed to an equivalent test that uses `_MSVC_LANG`.
> [...]
> As `_MSVC_LANG` is only defined in Visual Studio 2015 or newer it can't be used to test if Visual Studio 2012 or newer is used.

But we're not trying to test for VS 2012, we're trying to test for C++11 support. So couldn't this test be changed to:
```#if __cplusplus < 201103L && _MSVC_LANG < 201103L```

**Comment by TSlivede** [comment]
I don't exactly know which c++11 features are necessary for `cl2.hpp` but even Visual Studio 2012 did already [support some c++11 features](https://docs.microsoft.com/en-us/cpp/porting/visual-cpp-what-s-new-2003-through-2015#whats-new-for-c-in-visual-studio-2012) although it did not yet define `_MSVC_LANG`.

Therefore `_MSVC_LANG >= 201103L` can be evaluated false in Visual Studio although the necessary c++11 features are available. 

(This is under the assumption, that the c++11 features available in Visual Studio 2012 actually *are* enough to use `cl2.hpp`, otherwise the test is obviously flawed. As I don't have Visual Studio 2012 installed, I can't test if Visual Studio 2012 is sufficient.)

**Comment by jrprice** [comment]
I think we can be conservative here and just require full C++11 / Visual Studio 2013, if that simplifies these checks.

**Comment by TSlivede** [comment]
> just require full C++11 / Visual Studio 2013

Sadly Visual Studio 2013 was not fully compliant to c++11, although many c++11 features were present. (Even Visual Studio 2019 is AFAIK not yet really fully compliant to c++11.)

The test on [line 517](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/387174808abd743f869da345ccad6282b9fc845d/include/CL/cl2.hpp#L517)
```
#if (!defined(_MSC_VER) && __cplusplus < 201103L) || (defined(_MSC_VER) && _MSC_VER < 1700)
```
could be changed to this better readable version, that still checks for *c++11 or Visual Studio 2012*
```
#if !(__cplusplus >= 201103L || _MSC_VER >= 1700)
```
To test for *c++11 or Visual Studio 2013* the test would look like this (no difference in structure):
```
#if !(__cplusplus >= 201103L || _MSC_VER >= 1800)
```

> if that simplifies these checks

A further simplification (using [`_MSVC_LANG`](https://docs.microsoft.com/en-us/cpp/preprocessor/predefined-macros)) is only possible if we test for *c++11 or Visual Studio 2015 Update 3*:
```
#if !(__cplusplus >= 201103L || _MSVC_LANG >= 201103L)
```
This last option would be nice, because in this case the version check on [line 532](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/387174808abd743f869da345ccad6282b9fc845d/include/CL/cl2.hpp#L532) (the test, that this pull request was originally about) would no longer be necessary.

However this last option would also mean, that after this change `cl2.hpp` can no longer be used in Visual Studio 2012 and Visual Studio 2013...

**Comment by TSlivede** [comment]
>However, because of your question, I just thought, that
>```
>#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )
>```
>could possibly be changed into a test that uses `_MSC_VER`. That would be more consistent to the other test.

In case you decide, that [line 517](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/387174808abd743f869da345ccad6282b9fc845d/include/CL/cl2.hpp#L517) will be a test that uses `_MSC_VER`:

Yes, the proposed test for [line 532](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/387174808abd743f869da345ccad6282b9fc845d/include/CL/cl2.hpp#L532)
```
#if (__cplusplus >= 201103L || _MSVC_LANG >= 201103L )
```
can also be
```
#if (__cplusplus >= 201103L || _MSC_VER >= 1900)
```

jrprice:
OK, makes sense. Thanks for taking the time to explain the various options here, I appreciate it.

I think this PR is fine as is, so I'll merge it now.

Using `#if !(__cplusplus >= 201103L || _MSC_VER >= 1700)` for the [line 517](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/387174808abd743f869da345ccad6282b9fc845d/include/CL/cl2.hpp#L517) test is probably the best option since it doesn't change any requirements, so feel free to submit that as a follow up PR if you want to.

I'm not worried about changing anything in the old `cl.hpp` since it's deprecated.

PR #101, opened 2020-06-09
Add OpenCL 3.0 to the list of known versions

kpet:
Retain 2.2 as the default until 3.0 is final.


kpet:
@bashbaug Yes, but this would be done in a version later than 3.0 though. `CL_USE_DEPRECATED_OPENCL_3_0_APIS` would correspond to 3.0 API calls deprecated in a later version.

PR #102, opened 2020-06-09
adds a program build interface to compile for a specific device

### LINKED ISSUE #60: Please add specialization cl::Program::build(cl::Device&)
It is possible to create `cl::Context` using `cl::Device` instead of `std::vector<cl::Device>`, so it would be nice to use `cl::Program::build` in the same way.

BTW, this special constructor of `cl::Context` is not documented in opencl-cplusplus-1.2.pdf, please add it there.

bashbaug:
Fixes #60.

This complements the existing program build interface to compile for a vector of devices.

### DIFF
--- a/include/CL/cl2.hpp
+++ b/include/CL/cl2.hpp
@@ -2835,6 +2835,10 @@ public:
         }
     }
 
+    /*! \brief Constructs a context including a specific device.
+     *
+     *  Wraps clCreateContext().
+     */
     Context(
         const Device& device,
         cl_context_properties* properties = NULL,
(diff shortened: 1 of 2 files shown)


kpet:
Should we add a test?

kpet:
>I'm pretty sure the answer to this question is almost always "yes" smiley.

This was more of a rhetorical question. Seeing an approved PR adding new functionality without a test, I felt compelled :).

I think `buildErrHandler` and the unconditional gathering of build logs is useful functionality. Most users will want to see the build log after a failure. Having it directly packaged as part of the exception makes sense to me. Otherwise you need to catch build exceptions separately and then query the build log (or have a special case in the exception handler), this means more code for a very common pattern. Besides, performance really isn't a concern here.

Now I understand how that is a bit of a pain for testing. Maybe we could have utility testing functions to mirror what `buildErrHandler` and the typical calling code do?

bashbaug:
I think we should merge this PR as-is.  I would like to look at a utility testing function that mirrors `buildErrHandler`, but I don't expect to have a chance to look at this in the near future.

kpet:
Created https://github.com/KhronosGroup/OpenCL-CLHPP/issues/112 so we don't forget to add the utility function. Merging as agreed on the 2020/09/01 teleconference.

PR #104, opened 2020-06-12
Rename cl2.hpp to opencl.hpp

### LINKED ISSUE #25: Build cl.hpp for releases, not just cl2.hpp

kpet:
- Keep a cl2.hpp symlink for now for backwards compability
- Remove "CL2" from the name of examples
- Use opencl.hpp directly in tests and examples
- Remove a couple of unused variables in the trivial example

Fixes #25


README.md
```diff
@@ -1,17 +1,20 @@
...
 
 Components:
 
-  * `include/CL/cl2.hpp`:
+  * `include/CL/opencl.hpp`:
     The latest, maintained, version of the C++ bindings. It should work with all
     versions of OpenCL (including 1.x). This is what most users will want.
 
+  * `include/CL/cl2.hpp`:
+    Symbolic link to `opencl.hpp` for backwards compability.
```
jrprice, line 14: Do we need to make this an actual file that just includes `opencl.hpp` rather than a symlink, so that it also works on Windows? I don't have a strong opinion about this though.

jrprice: Although, if we made it an actual file, we could also add a `#warning` message that tells developers to start using the new header, which would help with the transition.

kpet: Don't symbolic links work on Windows these days? Anyway, I like your `#warning` idea, I'll go with that.

kpet:
Now rebased on top of https://github.com/KhronosGroup/OpenCL-CLHPP/pull/102. Tests passing. Merging since the rebase was simple.

PR #106, opened 2020-07-15
add initial support for OpenCL 3.0 platform and device queries

bashbaug:
Adds initial support for OpenCL 3.0 by adding the new OpenCL 3.0 platform and device queries.

A few notes:

* This needs to be merged after https://github.com/KhronosGroup/OpenCL-Headers/pull/96, since it relies on the device enqueue capabilities query and not the device enqueue boolean query.

* I'd specifically like feedback regarding interactions with the `cl_khr_extended_versioning` extension, since OpenCL 3.0 shares many of the queries with `cl_khr_extended_versioning`.  I've done this for now by separating the `cl_khr_extended_versioning` queries into "shared" queries that are also in OpenCL 3.0 and "unique" queries that aren't, but I'm open to suggestions to do this differently.

* An example application that uses these new queries can be found here: [newqueriespp](https://github.com/bashbaug/SimpleOpenCLSamples/tree/master/samples/00_newqueriespp).

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -18,7 +18,7 @@
  *
  *   \brief C++ bindings for OpenCL 1.0 (rev 48), OpenCL 1.1 (rev 33),
  *       OpenCL 1.2 (rev 15), OpenCL 2.0 (rev 29), OpenCL 2.1 (rev 17),
- *       and OpenCL 2.2 (V2.2-11).
+ *       OpenCL 2.2 (V2.2-11), and OpenCL 3.0 (V3.0.1-Provisional).
  *   \author Lee Howes and Bruce Merry
  *
  *   Derived from the OpenCL 1.x C++ bindings written by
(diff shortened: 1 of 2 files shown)


bashbaug:
Since https://github.com/KhronosGroup/OpenCL-Headers/pull/96 has been merged I re-triggered the CI builds and they are now passing.  I think this change is ready to go.

jrprice:
Not a thorough review, but can you add `CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED` as well so we don't forget? I think that's the only query that's been added since this PR was opened.

include/CL/cl2.hpp
```diff
@@ -18,7 +18,7 @@
  *
  *   \brief C++ bindings for OpenCL 1.0 (rev 48), OpenCL 1.1 (rev 33),
  *       OpenCL 1.2 (rev 15), OpenCL 2.0 (rev 29), OpenCL 2.1 (rev 17),
- *       and OpenCL 2.2 (V2.2-11).
+ *       OpenCL 2.2 (V2.2-11), and OpenCL 3.0 (V3.0.1-Provisional).
```
kpet, line 21: Do we want to mention the exact tag? It'll likely get out of date.

include/CL/cl2.hpp
```diff
@@ -1382,16 +1382,42 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
...
     F(cl_platform_info, CL_PLATFORM_NUMERIC_VERSION_KHR, cl_version_khr) \
     F(cl_platform_info, CL_PLATFORM_EXTENSIONS_WITH_VERSION_KHR, cl::vector<cl_name_version_khr>) \
     \
     F(cl_device_info, CL_DEVICE_NUMERIC_VERSION_KHR, cl_version_khr) \
-    F(cl_device_info, CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR, cl_version_khr) \
     F(cl_device_info, CL_DEVICE_EXTENSIONS_WITH_VERSION_KHR, cl::vector<cl_name_version_khr>) \
     F(cl_device_info, CL_DEVICE_ILS_WITH_VERSION_KHR, cl::vector<cl_name_version_khr>) \
     F(cl_device_info, CL_DEVICE_BUILT_IN_KERNELS_WITH_VERSION_KHR, cl::vector<cl_name_version_khr>)
 
+#define CL_HPP_PARAM_NAME_CL_KHR_EXTENDED_VERSIONING_UNIQUE_(F) \
```
kpet, line 1394: I'm wondering whether there wouldn't be a more descriptive name than `UNIQUE`. `KHRONLY`? `EXTENSION`? Maybe the above could be `SHARED_WITH_CORE`? Or a better name. I'm just thinking out loud really.

tests/test_cl2hpp.cpp
```diff
@@ -2429,7 +2447,10 @@ void testSetProgramSpecializationConstantPointer()
 #endif
 }
 
-// cl_khr_extended_versioning
+// OpenCL 3.0 and cl_khr_extended_versioning Queries
+
+// Note: This assumes the core enums, structures, and macros exactly match
+// the extension enums, structures, and macros.
```
kpet, line 2453: Should we have static asserts?

tests/test_cl2hpp.cpp
```diff
@@ -2484,8 +2523,13 @@ void testPlatformExtendedVersioning()
     TEST_ASSERT_EQUAL(extensions.size(), 1);
     TEST_ASSERT_EQUAL_HEX(extensions[0].version, CL_MAKE_VERSION_KHR(10, 11, 12));
     TEST_ASSERT_EQUAL_STRING(extensions[0].name, "cl_dummy_extension");
+#endif // CL_HPP_TARGET_OPENCL_VERSION < 300
```
kpet, line 2526: I think I'd rather we statically assert that the enums have the same value and use one rather than duplicate the test code.

tests/test_cl2hpp.cpp
```diff
@@ -2516,50 +2560,131 @@ static cl_int clGetDeviceInfo_extended_versioning(
...
+                "__opencl_c_feature",
+            },
+            {
+                CL_MAKE_VERSION(40, 43, 44),
+                "__opencl_c_fancy_feature",
+            },
+        };
+        if (param_value_size == sizeof(opencl_c_features) && param_value) {
+            cl_name_version* feature = static_cast<cl_name_version*>(param_value);
+            const int numFeatures = sizeof(opencl_c_features) / sizeof(opencl_c_features[0]);
```
kpet, line 2627: Do we not have an `ARRAY_SIZE` macro already? I couldn't find one, maybe it's a good excuse to introduce one.

tests/test_cl2hpp.cpp
```diff
@@ -2516,50 +2560,131 @@ static cl_int clGetDeviceInfo_extended_versioning(
...
+            for (int i = 0; i < numFeatures; i++) {
+                *feature = opencl_c_features[i];
+                ++feature;
```
kpet, line 2630: `feature[i] = ...`?

kpet (changes requested):
Looks good generally. I'm mostly concerned about the test code duplication.

include/CL/cl2.hpp
bashbaug: I'll just remove the tag for now.  It'll make OpenCL 3.0 look a little different than the prior OpenCL versions but maybe that's OK... and it's less error-prone anyhow.

include/CL/cl2.hpp
bashbaug: I changed to the following names, which I think are a bit more descriptive:

`CL_HPP_PARAM_NAME_CL_KHR_EXTENDED_VERSIONING_CL3_SHARED_` (for shared enums with OpenCL 3.0), and
`CL_HPP_PARAM_NAME_CL_KHR_EXTENDED_VERSIONING_KHRONLY_` (for the KHR-specific enums)

We can always change this later if we think of better names.

include/CL/cl2.hpp
jrprice: Now that the specs are unified, can't we just remove the all of tags and the "(rev XX)"s from all of the prior versions anyway?

tests/test_cl2hpp.cpp
bashbaug: Good idea - done.

tests/test_cl2hpp.cpp
bashbaug: I don't have a very strong opinion about this.  At least for the device queries, the KHR version tests `CL_DEVICE_OPENCL_C_NUMERIC_VERSION_KHR` and the 3.0 version doesn't.  Are you concerned about maintainability?  Test execution time?  Something else?

I've left this as-is for now.

tests/test_cl2hpp.cpp
bashbaug: Added.

tests/test_cl2hpp.cpp
bashbaug: Changed.

include/CL/cl2.hpp
bashbaug: I'm fine with this.  I'll remove all the "(rev XX)" information.

bashbaug:
I believe I have addressed all of the review feedback above.

The one item I haven't done yet is to add `CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED`.  I can do that now as part of this PR or in a follow-on PR.  Adding the query is easy but it'll be a bit more work to add a test.

bashbaug:
Actually, it looks like there aren't any tests for device queries like `CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED`, so I just added the query alongside the other new OpenCL 3.0 queries.  This is a gap we should fix but I think that can be done later.

kpet (approved):
LGTM

jrprice (approved):

PR #110, opened 2020-09-01
Fix enqueueMapSVM for cl::vector

paogor:
The proposed fix passes the container size _in bytes_ to clEnqueueSVMMap.


kpet (commented):
The fix looks good to me. Could you add a unit test please?

bashbaug (approved):
Thanks!  I'd missed the commit that added the unit test.

@kpet @jrprice any remaining concerns?

kpet (approved):
Looks fine. I'd missed the addition of the unit test as well.

Issue #80, opened 2019-11-06
Add support for creating a cl::CommandQueue with arbitrary properties


kpet:
Only properties defined using `CL_QUEUE_PROPERTIES` are currently supported. Arbitrary queue properties such as the ones added by vendor extensions aren't supported.

PR #113, opened 2020-09-14
Add partial support for cl_arm_scheduling_controls

kpet:
- Add support for device queries
- Introduce a generic function for clSetKernelExecInfo

Support for the command queue properties will be added via #80.


bashbaug (approved):
This generally looks good to me and I think this can be merged anytime.

Would we eventually want to add a templated version of `setExecInfo` similar to the `getInfo` functions though, so we could write code like this?

```c++
kernel.setExecInfo<CL_KERNEL_EXEC_INFO_SVM_FINE_GRAIN_SYSTEM>(CL_TRUE);
```

I think this would be less error-prone since we could enforce the argument type for a specific kernel exec info.

It's fine if we add this as a subsequent PR though.

kpet:
@bashbaug Good idea. I've added it here so I could add the type traits for cl_arm_scheduling_controls here. PTAL.

bashbaug (approved):
LGTM - thanks!

I do wonder if `void**` is the right type for `CL_KERNEL_EXEC_INFO_SVM_PTRS` (should it be a vector? should there be `const`?) but this is fine for now.

kpet:
>I do wonder if void** is the right type for CL_KERNEL_EXEC_INFO_SVM_PTRS

Agree this can be improved upon.

PR #118, opened 2020-10-27
guard the deprecated function clSetProgramReleaseCallback

bashbaug:
Since `clSetProgramReleaseCallback` was deprecated in OpenCL 3.0, we should guard calls to it using `CL_USE_DEPRECATED_OPENCL_2_2_APIS` to avoid a compiler warning in some cases.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -6794,6 +6794,7 @@ public:
     }
 
 #if CL_HPP_TARGET_OPENCL_VERSION >= 220
+#if defined(CL_USE_DEPRECATED_OPENCL_2_2_APIS)
     /*! \brief Registers a callback function to be called when destructors for
      *         program scope global variables are complete and before the
      *         program is released.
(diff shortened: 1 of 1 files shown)


kpet (approved):
Fine with this change as is as it's consistent with existing code but I can't understand why we use `CL_EXT_{PRE,SUF}FIX_.*_DEPRECATED` for deprecated core definitions and did not introduce `CL_API_{PRE,SUF}FIX_.*_DEPRECATED`.

bashbaug:
> I can't understand why we use `CL_EXT_{PRE,SUF}FIX_.*_DEPRECATED` for deprecated core definitions and did not introduce `CL_API_{PRE,SUF}FIX_.*_DEPRECATED`.

Yeah, there are several parts of these prefix and suffix defines that are confusing, see https://github.com/KhronosGroup/OpenCL-Headers/issues/95 and especially the comment [here](https://github.com/KhronosGroup/OpenCL-Headers/issues/95#issuecomment-663681382).

kpet:
I've +1'ed the comment. Good to see that is known/tracked.

kpet:
Straightforward, merging.

PR #122, opened 2020-12-10
switch from warning to pragma message for better portability

### LINKED ISSUE #121: `#warning` breaks MSVC builds including cl2.hpp
When using MSVC one gets the following error when trying to include cl2.hpp:
```
fatal error C1021: invalid preprocessor command 'warning'
```
Note that there is a solution on [SO](https://stackoverflow.com/a/279669).

bashbaug:
Fixes #121 

Some compiler do not support `#warning`, so switch to `#pragma message` instead.

### DIFF
--- a/include/CL/cl2.hpp
+++ b/include/CL/cl2.hpp
@@ -15,4 +15,4 @@
 //
 
 #include <CL/opencl.hpp>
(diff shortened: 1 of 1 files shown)


MathiasMagnus (approved):
shipit

PR #127, opened 2021-01-11
add missing const to context properties

bashbaug:
The properties that are passed to the `cl::Context` constructor should be const to match the call to `clCreateContext` or `clCreateContextFromType`.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -2851,7 +2851,7 @@ public:
      */
     Context(
         const vector<Device>& devices,
-        cl_context_properties* properties = NULL,
+        const cl_context_properties* properties = NULL,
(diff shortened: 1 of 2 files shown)


MathiasMagnus:
If we got two C++ API functions affected, how come only one test needed an update?

MathiasMagnus:
If we got two C++ API functions affected, how come only one test needed an update?

bashbaug:
I mostly wanted to ensure we had _some_ coverage for non-null (and const!) properties, but it was easy enough to add tests for the other two APIs also, so I think we have full test coverage now.  Thanks!

MathiasMagnus (approved):
Didn't mean to rain on your parade, but it does look better now.

jrprice (approved):

PR #128, opened 2021-02-16
Add copy constructor for ImageFormat

kpet:
Recent compilers require it because ImageFormat has a user-declared
assignment operator (-Wdeprecated-copy).


bashbaug (approved):
LGTM - I'll ignore the slight inconsistency vs. the implementation of the assignment operator 😄  (`*this = other` vs. `this->image_channel_data_type = other.image_channel_data_type, etc.`).

I suppose we could add a test for this too, but that's probably overkill.

jrprice (approved):

PR #129, opened 2021-02-16
Replace uses of CL_EXT_{PRE,SUF}FIX* with CL_API_{PRE,SUF}FIX*

kpet:
See KhronosGroup/OpenCL-Headers#153 for the
discussion.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -548,19 +548,19 @@
 
 // Define deprecated prefixes and suffixes to ensure compilation
 // in case they are not pre-defined
-#if !defined(CL_EXT_PREFIX__VERSION_1_1_DEPRECATED)
-#define CL_EXT_PREFIX__VERSION_1_1_DEPRECATED  
-#endif // #if !defined(CL_EXT_PREFIX__VERSION_1_1_DEPRECATED)
-#if !defined(CL_EXT_SUFFIX__VERSION_1_1_DEPRECATED)
-#define CL_EXT_SUFFIX__VERSION_1_1_DEPRECATED
-#endif // #if !defined(CL_EXT_PREFIX__VERSION_1_1_DEPRECATED)
+#if !defined(CL_API_PREFIX__VERSION_1_1_DEPRECATED)
+#define CL_API_PREFIX__VERSION_1_1_DEPRECATED  
+#endif // #if !defined(CL_API_PREFIX__VERSION_1_1_DEPRECATED)
+#if !defined(CL_API_SUFFIX__VERSION_1_1_DEPRECATED)
+#define CL_API_SUFFIX__VERSION_1_1_DEPRECATED
+#endif // #if !defined(CL_API_PREFIX__VERSION_1_1_DEPRECATED)
 
-#if !defined(CL_EXT_PREFIX__VERSION_1_2_DEPRECATED)
-#define CL_EXT_PREFIX__VERSION_1_2_DEPRECATED  
-#endif // #if !defined(CL_EXT_PREFIX__VERSION_1_2_DEPRECATED)
-#if !defined(CL_EXT_SUFFIX__VERSION_1_2_DEPRECATED)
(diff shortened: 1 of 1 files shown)


jrprice (approved):

kpet:
The CI will fail until https://github.com/KhronosGroup/OpenCL-Headers/pull/158 gets in.

include/CL/opencl.hpp
```diff
@@ -548,19 +548,19 @@
...
+#if !defined(CL_API_SUFFIX__VERSION_1_1_DEPRECATED)
+#define CL_API_SUFFIX__VERSION_1_1_DEPRECATED
+#endif // #if !defined(CL_API_PREFIX__VERSION_1_1_DEPRECATED)
+
+#if !defined(CL_API_PREFIX__VERSION_1_2_DEPRECATED)
+#define CL_API_PREFIX__VERSION_1_2_DEPRECATED  
+#endif // #if !defined(CL_API_PREFIX__VERSION_1_2_DEPRECATED)
+#if !defined(CL_API_SUFFIX__VERSION_1_2_DEPRECATED)
+#define CL_API_SUFFIX__VERSION_1_2_DEPRECATED
+#endif // #if !defined(CL_API_PREFIX__VERSION_1_2_DEPRECATED)
```
bashbaug, line 563: Could we add the `VERSION_2_2_DEPRECATED` prefixes and suffixes here in case they are not pre-defined?  This would enable CI to pass even without the header updates, which is good because it's very possible that the the headers and the C++ bindings could be out-of-sync for a bit.

bashbaug (changes requested):
Can we find a way to make this work even if the definitions for the headers and C++ bindings are out-of-sync for a bit?  One idea is mentioned below, which will allow the newer C++ bindings to work with older headers (I think).

include/CL/opencl.hpp
bashbaug: I'm remembering now that I added the 2.2 deprecation guards (https://github.com/KhronosGroup/OpenCL-CLHPP/pull/118).  I should have added the "in case they are not pre-defined" 2.2 prefixes and suffixes here at the same time - sorry about that!

This raises an interesting question though: should the functions in the C++ bindings _also_ include the deprecation prefixes and suffixes?  Or, is this overkill, since the OpenCL API calls will already include a deprecation message?

bashbaug: More to the point.  In the C++ bindings we have (this is a nice short example, there are other places that use the same pattern):

```c++
#if defined(CL_USE_DEPRECATED_OPENCL_1_1_APIS)
/**
 * Unload the OpenCL compiler.
 * \note Deprecated for OpenCL 1.2. Use Platform::unloadCompiler instead.
 */
inline CL_API_PREFIX__VERSION_1_1_DEPRECATED cl_int
UnloadCompiler() CL_API_SUFFIX__VERSION_1_1_DEPRECATED;
inline cl_int
UnloadCompiler()
{
    return ::clUnloadCompiler();
}
#endif // #if defined(CL_USE_DEPRECATED_OPENCL_1_1_APIS)
```

When we have `CL_USE_DEPRECATED_OPENCL_1_1_APIS` defined, `CL_API_PREFIX__VERSION_1_1_DEPRECATED` is always defined to nothing, so it's not adding anything.  Even if we didn't have the guard for `CL_USE_DEPRECATED_OPENCL_1_1_APIS`, we will already get a warning that `clUnloadCompiler` was declared deprecated, so do we really need a deprecation warning on `UnloadCompiler` also?

bashbaug (approved):
@kpet can you please take a look at the comments above?

I'm fine with these changes going in as-is since the build is currently broken, but we should definitely find ways to make this more resilient.

kpet:
I did have a look and I agree there's a discussion to be had. I also agree we should get this in to fix the build ASAP. I've created as a follow-up issue (https://github.com/KhronosGroup/OpenCL-CLHPP/issues/131). Merging this since you've now approved the change.

PR #130, opened 2021-02-24
Fix cl2.hpp so it compiles for both 1.2 and 2.0

ouakheli:
The compiler complains about missing braces when compiling for 2.0, and
complains about "too many braces" for 1.2, because cl_image_desc has a
union in 2.0, but just a cl_mem in 1.2.

Set only needed members of cl_image_desc struct.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -4478,12 +4478,11 @@ public:
         cl_int* err = NULL)
     {
         cl_int error;
-        cl_image_desc desc =
-        {
-            CL_MEM_OBJECT_IMAGE1D,
-            width,
-            0, 0, 0, 0, 0, 0, 0, 0
-        };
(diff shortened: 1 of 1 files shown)


bashbaug:
Thank you for your contribution!  The travis build failure is not caused by your change - see #129.

bashbaug (approved):
LGTM, this is a nice improvement.

Do we need to enhance our CI coverage to catch issues like this in the future?

It'd be great to get another quick review from @jrprice or @kpet  before merging.

jrprice (approved):

jrprice:
> Do we need to enhance our CI coverage to catch issues like this in the future?

Yes, wouldn't hurt. I've opened https://github.com/KhronosGroup/OpenCL-CLHPP/issues/133 for this.

Issue #131, opened 2021-02-24
Revisit how we handle deprecated functions


kpet:
Do we really need the bindings to be declared as "deprecated" when the underlying CL function already is?

See discussion on https://github.com/KhronosGroup/OpenCL-CLHPP/pull/129

bashbaug:
We should revisit how these C++ bindings handle deprecated enums too - see related review comment https://github.com/KhronosGroup/OpenCL-CLHPP/pull/136#pullrequestreview-644858031.

We currently have a small code block to handle enums that were deprecated in OpenCL 2.0:

```c++
// Flags deprecated in OpenCL 2.0
#define CL_HPP_PARAM_NAME_INFO_1_0_DEPRECATED_IN_2_0_(F) \
    F(cl_device_info, CL_DEVICE_QUEUE_PROPERTIES, cl_command_queue_properties)

#define CL_HPP_PARAM_NAME_INFO_1_1_DEPRECATED_IN_2_0_(F) \
    F(cl_device_info, CL_DEVICE_HOST_UNIFIED_MEMORY, cl_bool)

#define CL_HPP_PARAM_NAME_INFO_1_2_DEPRECATED_IN_2_0_(F) \
    F(cl_image_info, CL_IMAGE_BUFFER, cl::Buffer)

// Include deprecated query flags based on versions
// Only include deprecated 1.0 flags if 2.0 not active as there is an enum clash
#if CL_HPP_TARGET_OPENCL_VERSION > 100 && CL_HPP_MINIMUM_OPENCL_VERSION < 200 && CL_HPP_TARGET_OPENCL_VERSION < 200
CL_HPP_PARAM_NAME_INFO_1_0_DEPRECATED_IN_2_0_(CL_HPP_DECLARE_PARAM_TRAITS_)
#endif // CL_HPP_MINIMUM_OPENCL_VERSION < 110
#if CL_HPP_TARGET_OPENCL_VERSION > 110 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
CL_HPP_PARAM_NAME_INFO_1_1_DEPRECATED_IN_2_0_(CL_HPP_DECLARE_PARAM_TRAITS_)
#endif // CL_HPP_MINIMUM_OPENCL_VERSION < 120
#if CL_HPP_TARGET_OPENCL_VERSION > 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
[...]

PR #136, opened 2021-04-24
add missing getInfo query mappings

### LINKED ISSUE #132: Add missing clGetDeviceInfo queries
The CL_DEVICE_PRINTF_BUFFER_SIZE (missing before version 1.2) device info query is not listed in [clGetInfoHelper]( https://github.com/KhronosGroup/OpenCL-CLHPP/blob/2ba05c4d2173db2536f5a157ff94afa8bb355016/include/CL/opencl.hpp#L1145).

Need to add this one and audit the remaining device info queries for completeness.
.

bashbaug:
This PR adds the missing query mapping for `CL_DEVICE_PRINTF_BUFFER_SIZE` mentioned by #132, adds a few other query mappings that were missing, and fixes a few other query mappings that were incorrect.

Fixes #132.  Tagging @jlewis-austin for possible review.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -1317,14 +1317,20 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     F(cl_kernel_arg_info, CL_KERNEL_ARG_NAME, string) \
     F(cl_kernel_arg_info, CL_KERNEL_ARG_TYPE_QUALIFIER, cl_kernel_arg_type_qualifier) \
     \
+    F(cl_kernel_work_group_info, CL_KERNEL_GLOBAL_WORK_SIZE, cl::detail::size_t_array) \
+    \
+    F(cl_device_info, CL_DEVICE_LINKER_AVAILABLE, cl_bool) \
+    F(cl_device_info, CL_DEVICE_IMAGE_MAX_BUFFER_SIZE, size_type) \
+    F(cl_device_info, CL_DEVICE_IMAGE_MAX_ARRAY_SIZE, size_type) \
     F(cl_device_info, CL_DEVICE_PARENT_DEVICE, cl::Device) \
     F(cl_device_info, CL_DEVICE_PARTITION_MAX_SUB_DEVICES, cl_uint) \
     F(cl_device_info, CL_DEVICE_PARTITION_PROPERTIES, cl::vector<cl_device_partition_property>) \
     F(cl_device_info, CL_DEVICE_PARTITION_TYPE, cl::vector<cl_device_partition_property>)  \
     F(cl_device_info, CL_DEVICE_REFERENCE_COUNT, cl_uint) \
-    F(cl_device_info, CL_DEVICE_PREFERRED_INTEROP_USER_SYNC, size_type) \
+    F(cl_device_info, CL_DEVICE_PREFERRED_INTEROP_USER_SYNC, cl_bool) \
     F(cl_device_info, CL_DEVICE_PARTITION_AFFINITY_DOMAIN, cl_device_affinity_domain) \
     F(cl_device_info, CL_DEVICE_BUILT_IN_KERNELS, string) \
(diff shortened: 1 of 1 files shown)


include/CL/opencl.hpp
```diff
@@ -1364,15 +1375,15 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
 #define CL_HPP_PARAM_NAME_INFO_2_1_(F) \
     F(cl_platform_info, CL_PLATFORM_HOST_TIMER_RESOLUTION, size_type) \
```
jlewis-austin, line 1376: This is cl_ulong in the spec

include/CL/opencl.hpp
```diff
@@ -1364,15 +1375,15 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
...
     F(cl_program_info, CL_PROGRAM_IL, cl::vector<unsigned char>) \
```
jlewis-austin, line 1377: Not necessarily a problem, but I'm curious about the choice of cl::vector<unsigned char> vs string. Most of the queries use string where char[] is specified, but I see the following and maybe a couple others use vectors:
CL_PROGRAM_IL_KHR
CL_PROGRAM_IL
CL_PROGRAM_BINARY_SIZES

jlewis-austin (commented):
The changes LGTM. I had a few comments based on a review of all the queries; I'm fine with addressing them here or deferring to another PR. Comments it wouldn't let me inline:

- A few queries are deprecated in 3.0 but aren't marked as such:

CL_DEVICE_OPENCL_C_VERSION
CL_PROGRAM_SCOPE_GLOBAL_CTORS_PRESENT
CL_PROGRAM_SCOPE_GLOBAL_DTORS_PRESENT

- CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE is included in the 1.1 section here, but is not marked as missing before 1.1 in the 3.0 spec. It looks like it was introduced in 1.1. I can make a spec issue if that's right.

include/CL/opencl.hpp
bashbaug: Thanks, fixed!

include/CL/opencl.hpp
bashbaug: It's because these queries return binary data that isn't necessarily guaranteed to be null-terminated.

Interestingly the spec for [CL_PROGRAM_IL](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#CL_PROGRAM_IL) says the return value is a `char[]`, whereas the query for CL_PROGRAM_IL_KHR returns an `unsigned char[]`.  In practice, I'm not sure that it matters, but the difference is a little odd, and these C++ bindings return a `cl::vector<unsigned char>` in both cases for whatever that's worth.

bashbaug:
Thanks @jlewis-austin!

> * CL_KERNEL_PREFERRED_WORK_GROUP_SIZE_MULTIPLE is included in the 1.1 section here, but is not marked as missing before 1.1 in the 3.0 spec. It looks like it was introduced in 1.1. I can make a spec issue if that's right.

I noticed this too and filed a bug against the headers: https://github.com/KhronosGroup/OpenCL-Headers/issues/167 😄 

When we fix the headers we should fix the XML file also, which should add the right text to the spec also.

> * A few queries are deprecated in 3.0 but aren't marked as such...

I'll take a look at this tomorrow, but if this doesn't make it into this PR we can add it as a subsequent PR.  Thanks!

bashbaug:
> * A few queries are deprecated in 3.0 but aren't marked as such...

I took a look and I'd prefer to handle deprecated enums separately.  I think this could be a much bigger change.  I added a comment to a related issue regarding deprecated functionality https://github.com/KhronosGroup/OpenCL-CLHPP/issues/131#issuecomment-827314259 so we don't lose track of it.  Thanks again!

jlewis-austin (approved):
Thanks, LGTM!

jrprice (approved):

PR #137, opened 2021-06-21
add queries for cl_khr_integer_dot_product

bashbaug:
Adds query support for the `cl_khr_integer_dot_product` extension.

See: https://github.com/KhronosGroup/OpenCL-Docs/pull/627

This PR won't break anything as-is, but it will require the corresponding header changes in order to use these changes.

See: https://github.com/KhronosGroup/OpenCL-Headers/pull/170


kpet (approved):
LGTM, thanks! Would be nice to have a test but fine with the PR as is.

bashbaug:
Thanks!  I filed #138 to track the question of device info queries more broadly.

PR #139, opened 2021-08-05
Fix unreasonable error handling in opencl.hpp

TianZerL:
This PR fixes two potential bracket mismatches in opencl.hpp.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -6424,9 +6424,8 @@ public:
         static PFN_clCreateProgramWithILKHR pfn_clCreateProgramWithILKHR = NULL;
         CL_HPP_INIT_CL_EXT_FCN_PTR_(clCreateProgramWithILKHR);
 
-        return detail::errHandler(
-            pfn_clCreateProgramWithILKHR(
-                context(), static_cast<const void*>(IL.data()), IL.size(), &error));
+        object_ = pfn_clCreateProgramWithILKHR(
+                context(), static_cast<const void*>(IL.data()), IL.size(), &error);
 
 #endif // #if CL_HPP_TARGET_OPENCL_VERSION >= 210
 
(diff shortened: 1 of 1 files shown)


alycm (approved):
Thank you this submission, the code is clearly wrong.

I checked why this had not been spotted before and it is because these lines are only compiled when `CL_HPP_TARGET_OPENCL_VERSION==200 && defined(CL_HPP_USE_IL_KHR)`.  Which is untested.  I'll file an issue about that.

alycm:
Although the code as-is is clearly wrong, and this PR makes the code less wrong, the lines that are being changed are still very wrong.

The pass a `cl_program` to a error checking function that is expecting a `cl_int`, and then return a `cl_int` from a constructor (which of course can't return a value at all).

include/CL/opencl.hpp
```diff
@@ -6426,7 +6426,7 @@ class Program : public detail::Wrapper<cl_program>
 
         return detail::errHandler(
             pfn_clCreateProgramWithILKHR(
-                context(), static_cast<const void*>(IL.data()), IL.size(), &error);
+                context(), static_cast<const void*>(IL.data()), IL.size(), &error));
```
alycm, line 6429: I think these lines should be

```c++
        object_ = pfn_clCreateProgramWithILKHR(
                 context(), static_cast<const void*>(IL.data()), IL.size(), &error);
```

include/CL/opencl.hpp
```diff
@@ -6480,7 +6480,7 @@ class Program : public detail::Wrapper<cl_program>
 
         return detail::errHandler(
             pfn_clCreateProgramWithILKHR(
-            context(), static_cast<const void*>(IL.data()), IL.size(), &error);
+            context(), static_cast<const void*>(IL.data()), IL.size(), &error));
```
alycm, line 6483: I think these lines should be

```c++
        object_ = pfn_clCreateProgramWithILKHR(
             context(), static_cast<const void*>(IL.data()), IL.size(), &error);
```

alycm (changes requested):

alycm:
I very quickly extended the testing to cover this case, and a few others, here: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/140

That of course can't merge until this PR does, as the tests do not build without this fix.

include/CL/opencl.hpp
TianZerL: Yes, it is incorrect to handle errors in this branch. I will modify the PR.

alycm (approved):
Thank you for making the changes!  I've confirmed that this builds with https://github.com/KhronosGroup/OpenCL-CLHPP/pull/140.

PR #140, opened 2021-08-05
Test some of the possible defines

alycm:
Fix the test suite so that stubs always set the return error code, otherwise
`CL_HPP_ENABLE_EXCEPTIONS` causes uncaught exceptions to be thrown.

The `CL_HPP_NO_STD_FOO` defines are not tested, because the header won't build
unless an alternative implementation is provided


alycm:
Note that these tests won't build until the issue identied in https://github.com/KhronosGroup/OpenCL-CLHPP/pull/139 is fixed.

This increases the number of test from 5 to 32, but it still takes under a minute to build and seconds to run the test.

I'm not familiar with any of these options beyond inferring what they do from their name, so I don't know with certainty that these configs are meaningful to test.

bashbaug (approved):
Changes LGTM, though we need to merge this after #139.

The C++ bindings currently have special defines and behavior for:

* `cl_ext_device_fission` (via `CL_HPP_USE_CL_DEVICE_FISSION`)
* `cl_khr_d3d10_sharing` (via `CL_HPP_USE_DX_INTEROP`)
* `cl_khr_subgroups` (via `CL_HPP_USE_CL_SUB_GROUPS_KHR`)
* `cl_khr_image2d_from_buffer` (via `CL_HPP_USE_CL_IMAGE2D_FROM_BUFFER_KHR`)

That's a rather odd assortment of extensions, and it's rather odd that we need to use a specific define to enable them.  Could we find a way to eanble this extension functionality by default, without needing a special define?

To be clear, this should NOT hold up this PR, and if we decide to do this (or even that more discussion is needed) I'll open a separate issue for tracking.  Thanks!

alycm:
I think that `CL_HPP_USE_IL_KHR` could be added to that list too.

I agree that needing to enable those extensions is odd.  It doesn't make much sense in any context were compiling is independent of deployment, i.e. if you don't know what device you will be running on you can't know at compile time if the extension will exist anyway.

This PR doesn't test `CL_HPP_USE_CL_IMAGE2D_FROM_BUFFER_KHR` or `CL_HPP_USE_DX_INTEROP`.  I've also realized that this PR tests `CL_HPP_ENABLE_DEVICE_FISSION` as described in the header description, but the true define is `CL_HPP_USE_CL_DEVICE_FISSION`.

I'll fix those things in a future PR (the tests and the header comment).  Or it becomes irrelevant if we drop the defines.

bashbaug:
I created #146 to track the extension define removal.  Merging.

PR #145, opened 2021-08-31
Fix cl::enqueueMapSVM for cl::vector and cl::pointer

ybh1998:
Fix the missing reference symbol for cl::enqueueMapSVM. Otherwise vector will be coppied and enqueueMapSVM will map to a different pointer address. Added tests.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -9320,7 +9320,7 @@ inline cl_int enqueueMapSVM(
  */
 template<typename T, class D>
 inline cl_int enqueueMapSVM(
-    cl::pointer<T, D> ptr,
+    cl::pointer<T, D> &ptr,
     cl_bool blocking,
     cl_map_flags flags,
     size_type size,
@@ -9344,7 +9344,7 @@ inline cl_int enqueueMapSVM(
  */
 template<typename T, class Alloc>
 inline cl_int enqueueMapSVM(
-    cl::vector<T, Alloc> container,
+    cl::vector<T, Alloc> &container,
     cl_bool blocking,
     cl_map_flags flags,
     const vector<Event>* events = NULL,
--- a/tests/cmock.yml
+++ b/tests/cmock.yml
@@ -7,3 +7,5 @@
     - CL_API_CALL
   :plugins:
     - :callback
+  :treat_as:
+    'void*': 'PTR'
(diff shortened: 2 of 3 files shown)


bashbaug (commented):
Thank you for your contribution, and especially for adding a test for this case!

Your fix looks good and the automated tests are passing but I'm getting a segfault when I run the test on Windows.  I'd like to hold off merging this change until I figure out what's going on.  I'm looking at it now so hopefully this shouldn't take too long...

bashbaug:
OK, I know a little more what's happening.  Short summary:

* Linux seems to be working fine for debug and release builds.
* Windows (MSVC) is working fine for release builds.
* Windows is NOT working fine for debug builds.
* The reason Windows debug builds are not working is because even an empty vector is calling into the allocator.  Since the new test is not expecting calls to clSVMAlloc (and any associated retains and releases for the context) the test fails.

I think we can "fix" this by setting `_ITERATOR_DEBUG_LEVEL` equal to `0`, see [link](https://docs.microsoft.com/en-us/cpp/standard-library/iterator-debug-level?view=msvc-160).

I'd be fine merging this as-is with an issue to fix this, and I'll create a PR to set `_ITERATOR_DEBUG_LEVEL` shortly.  Sound reasonable?

Tagging @jrprice and @kpet for awareness.

ybh1998:
I think maybe it is better to use an initialized `std::vector` to test this fix? Just like the `testEnqueueMapSVM`.
@bashbaug could you please test the updated commit on Windows (MSVC)?

bashbaug:
Thank you, your latest commit runs fine on Windows (MSVC).... but now the test isn't failing without your fix!

I traced this to the default CMock behavior, which treats `void*` arguments as `HEX8ARRAY`, meaning it checks the contents of memory instead of the pointer value.

https://github.com/ThrowTheSwitch/CMock/blob/master/docs/CMock_Summary.md

We can fix this with a change to the cmock YAML file, by adding:

```yaml
  :treat_as:
    'void*' : PTR
```

I think that's the right thing to do and all tests are passing with this change when I run them on my system.  We could add this separately or as part of your PR.  Let me know what you'd prefer to do.  Thanks again!

ybh1998:
PR updated. I didn't even notice this problem. Thank you very much! @bashbaug

bashbaug (approved):
Thank you!  Merging.

Issue #146, opened 2021-09-02
revisit handling for extensions


bashbaug:
Creating this issue in response to a review comment: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/140#pullrequestreview-725536988

We currently have a handful of defines in the C++ bindings for extensions.  Are these needed?  If not, we can remove the defines to simplify testing (this is preferred)?  If so, do we need to add defines for other extensions?

PR #147, opened 2021-09-06
Test two additional defines and update descriptions in header

alycm:
Newly tested defines:
* `CL_HPP_USE_CL_DEVICE_FISSION` (corrected from `CL_HPP_ENABLE_DEVICE_FISSION`)
* `CL_HPP_USE_CL_IMAGE2D_FROM_BUFFER_KHR`

The `CL_HPP_NO_STD_FOO` defines are not tested, because the header won't build
unless an alternative implementation is provided.  `The CL_HPP_USE_DX_INTEROP`
define is not tested because it does not work on platforms without DirectX
headers.

Issue #146 could make this irrelevant, but easy to test these defines in the
meantime.


bashbaug (approved):

bashbaug:
I just created #148 which should make these changes unnecessary, but these changes are straightforward, fix the documentation, and increase test coverage, so it seems like a good idea to merge them anyhow.

PR #148, opened 2021-09-08
remove extension defines

### LINKED ISSUE #146: revisit handling for extensions

bashbaug:
Fixes #146

This PR removes many of the extension defines from the C++ bindings and enables support for extensions based on defines in the OpenCL headers instead.

Specifically, the following extensions defines are removed:

* `CL_HPP_USE_CL_DEVICE_FISSION` (for `cl_ext_device_fission`)
* `CL_HPP_USE_CL_SUB_GROUPS_KHR` (for `cl_khr_subgroups`)
* `CL_HPP_USE_IL_KHR` (for `cl_khr_il_program`)
* `CL_HPP_USE_CL_IMAGE2D_FROM_BUFFER_KHR` (for `cl_khr_image2d_from_buffer`)

The extension define for `CL_HPP_USE_DX_INTEROP` (for `cl_khr_d3d10_sharing`) has not been removed since the DX interop APIs require including the DX-specific header and hence should be enabled conditionally.

Some implementation notes:

* We can differentiate between the core OpenCL 1.2 sub-device creation and the `cl_ext_device_fission` sub-device creation based on the function overload, since `cl_device_partition_property` is a different type than `cl_device_partition_property_ext` (`intptr_t` vs. `cl_ulong`).
* For subgroups, we'll try the core OpenCL 2.1 APIs first if we're compiling for OpenCL 2.1 or newer and the platform is an OpenCL 2.1 platform, and we'll fall back to the extension APIs if the core APIs are unsupported or return an error.
* Similarly, for IL programs, we'll try the core OpenCL 2.1 APIs first and only fall back to the extension APIs if the core APIs are unsupported or return an error.
* We do not currently have a runtime check for the extension.  I did add a check whether the extension API function pointer was successfully retrieved, though.
* Because `cl_khr_image2d_from_buffer` uses the same `clCreateImage` API for both the core and extension version there is no need to try both.  Additionally, there is no runtime check, and the API call will either succeed or return an error.

I think all of this is reasonble and correct but please double-check my reasoning!

A note about testing:

* I added tests for the different codepaths for `cl_khr_subgroups` and `cl_khr_il_program`, but this admittedly isn't perfect: because the C++ bindings store the extension API function pointers in static variables they are "stateful" and therefore must run in a specific order.  We should rethink how extensions are handled to improve robustness, especially because the C++ bindings currently will not work for many extension APIs from multiple platforms.


bashbaug:
I've found a few issues with this PR that I need to fix:

1. `CL_HPP_INIT_CL_EXT_FCN_PTR_` uses `clGetExtensionFunctionAddress` to get the function pointer, but this was declared deprecated, leading to a deprecation warning when compiling for some combinations of the minimum and target version of OpenCL.
2. I'm using `getDevicePlatformVersion` to check whether the core API or extension API should be called, but this is only defined for some combinations of the minimum and target version of OpenCL.

I'll mark this PR as ready for review when I have solved these issues.

include/CL/opencl.hpp
```diff
@@ -2396,35 +2385,39 @@ class Device : public detail::Wrapper<cl_device_id>
...
+                devices->resize(ids.size());
+
+                // Assign to param, constructing with retain behaviour
+                // to correctly capture each underlying CL object
+                for (size_type i = 0; i < ids.size(); i++) {
+                    // We do not need to retain because this device is being created 
+                    // by the runtime
+                    (*devices)[i] = Device(ids[i], false);
+                }
             }
```
MathiasMagnus, line 2415: IMHO this function is unfortunate in multiple ways, some of which are breaking so I won't list. However:
- if the user doesn't supply a vector, then the entire function is meaningless and we should shortcircuit the entire thing and not leak subdevice handles.
- the SDK utility library introduced new error codes for cases when the wrapper itself is misused. This function really has no meaning when a `nullptr` is provided by the user and we should likely signal that. Whether we repurpose an error code from [`clCreateSubDevices`](https://www.khronos.org/registry/OpenCL/specs/3.0-unified/html/OpenCL_API.html#clCreateSubDevices) or start making our own, it's up to us. Silently leaking resources seems wrong.

I'm fine deferring this into a separate issue.

include/CL/opencl.hpp
```diff
@@ -2396,35 +2385,39 @@ class Device : public detail::Wrapper<cl_device_id>
...
+            if (devices) {
+                devices->resize(ids.size());
+
+                // Assign to param, constructing with retain behaviour
+                // to correctly capture each underlying CL object
+                for (size_type i = 0; i < ids.size(); i++) {
+                    // We do not need to retain because this device is being created 
+                    // by the runtime
+                    (*devices)[i] = Device(ids[i], false);
+                }
```
MathiasMagnus, line 2414: In the spirit of no raw loops. Mild bikeshedding, feel free to discard if more trouble than use.
```suggestion
                devices->reserve(ids.size());

                // Assign to param, constructing with retain behaviour
                // to correctly capture each underlying CL object
                std::transform(ids.cbegin(), ids.cend(), std::back_inserter(devices), [](const cl_device_id& id) {
                    // We do not need to retain because this device is being created 
                    // by the runtime
                    return Device(id, false);
                });
```

MathiasMagnus: Mind you, this requires `algorithm`. `iterator` is already included.

MathiasMagnus (approved):
Overall looks good to me. Minor stylistic suggestion (can be downvoted) and a suggestion to defer a fix which may cover multiple functions, not just the changed one.

PR #151, opened 2021-10-14
update the C++ bindings code example

### LINKED ISSUE #149: Update example code
As novice open-cl-clhpp user i was delighted to see the example code highlighting usage etc.. As this is from the official open-cl-clhpp and in the source code i hoped it would be showing idiomatic usage and be up-to-date. 

However it took me some effort and frustration to get it working. In order to give other new users a better experience i would propose the following changes.

1) Have the program check for open_cl version >= 200
I ran the program on my Intel i5-8400 with integrated GPU which reports itself as Platform Version OpenCL 3.0. So it refused to run as this is not opencl 2.0

2) Mention (or have the program check for) minimum compile options -pthread -lOpenCL
This may be obvious to most of you, but I never thought of compiling with -pthread. On my system this gave very uninformative 'system error' and just quit. (turned out makedefault uses std::runonce which needs threads)

3) The program tries to read an output2 which was never defined
Probably original code wanted to show reading from SVM and device memory but one got removed.
This was easy but still i think the code should run without needing to sort this out and fixing it.

As to what the program showcases and how is a matter of opinion. To me it was very useful as is. I do have some questions though
 - shouldn't there be an "if(get_global_id(0)==0)" in kernel2 before enqueuing the new commands? Now all 16 workers will enqueue a kernel that writes to the third and last quarter of the array. So these values each get written 16 times.
 - Following on the previous point: why doesn't this lead to problems (accessing the same array value from multiple workers)? I tried using printf() statements to follow but they all come out in (ascending work-item) order.

Link to the doxygen page with example code
[http://github.khronos.org/OpenCL-CLHPP/](url)

bashbaug:
fixes #149 

Updates the code example in both the cpp source and the header itself (for doxygen).

We'll need to regenerate and update the docs after this PR is merged.

### DIFF
--- a/examples/src/headerexample.cpp
+++ b/examples/src/headerexample.cpp
@@ -11,28 +11,30 @@ const int numElements = 32;
 
 int main(void)
 {
-    // Filter for a 2.0 platform and set it as the default
+    // Filter for a 2.0 or newer platform and set it as the default
     std::vector<cl::Platform> platforms;
     cl::Platform::get(&platforms);
     cl::Platform plat;
     for (auto &p : platforms) {
         std::string platver = p.getInfo<CL_PLATFORM_VERSION>();
-        if (platver.find("OpenCL 2.") != std::string::npos) {
+        if (platver.find("OpenCL 2.") != std::string::npos ||
+            platver.find("OpenCL 3.") != std::string::npos) {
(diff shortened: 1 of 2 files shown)


jrprice (approved):

alycm (approved):
LGTM, this is the second review, merging.

Issue #158, opened 2021-11-05
6891:5: error: ‘CL_API_PREFIX__VERSION_2_2_DEPRECATED’ does not name a type; did you mean ‘CL_API_PREFIX__VERSION_1_2_DEPRECATED’?


Greatpanc:
[  0%] Building CXX object examples/src/CMakeFiles/headerexample.dir/headerexample.cpp.o
[  0%] Linking CXX executable headerexample
[  0%] Built target headerexample
[  0%] Building CXX object examples/src/CMakeFiles/trivialSizeTCompat.dir/trivialSizeTCompat.cpp.o
[  1%] Linking CXX executable trivialSizeTCompat
[  1%] Built target trivialSizeTCompat
[  1%] Building CXX object examples/src/CMakeFiles/trivial.dir/trivial.cpp.o
[  1%] Linking CXX executable trivial
[  1%] Built target trivial
[  1%] Generating test runner
[  1%] Stripping defines from cl.h
[  2%] Generating mocks
Creating mock for cl...
[  2%] Building C object tests/CMakeFiles/test_openclhpp_300_CL_HPP_USE_CL_SUB_GROUPS_KHR.dir/test_openclhpp_Runner.c.o
[  2%] Building CXX object tests/CMakeFiles/test_openclhpp_300_CL_HPP_USE_CL_SUB_GROUPS_KHR.dir/test_openclhpp.cpp.o
In file included from /home/greatpan/workspace/OpenCL-CLHPP/tests/test_openclhpp.cpp:5:
/home/greatpan/workspace/OpenCL-CLHPP/include/CL/opencl.hpp:6891:5: error: ‘CL_API_PREFIX__VERSION_2_2_DEPRECATED’ does not name a type; did you mean ‘CL_API_PREFIX__VERSION_1_2_DEPRECATED’?
 6891 |     CL_API_PREFIX__VERSION_2_2_DEPRECATED cl_int setReleaseCallback(
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
      |     CL_API_PREFIX__VERSION_1_2_DEPRECATED
/home/greatpan/workspace/OpenCL-CLHPP/tests/test_openclhpp.cpp: In function ‘void testSetProgramReleaseCallback()’:
[...]

PR #159, opened 2021-11-07
Added missing definitions for deprecated prefixes and suffixes for ve…

pkeir:
…rsion 2.2. This will resolve issue #158 .


bashbaug (approved):
LGTM, thank you!

Should we create an issue to add similar blocks for the 1.0, 2.0, and 2.1 deprecation defines?  They aren't currently used as far as I can see, but this could prevent a similar bug in the future.

pkeir:
No problem. Yes, I would say so, for the reason you mention. In each future release it might become an almost mechanical addition.

bashbaug:
I created #161 to track definitions for the other OpenCL versions.  Merging - thank you again!

PR #164, opened 2021-11-22
Add support for const iterators to the 3 Buffer constructors.

pkeir:
Allow const iterator arguments for the 3 Buffer constructors which have iterator parameters. Code such as the following can now compile:
```
cl::Buffer buf_b1(data_a.cbegin(), data_a.cend(), true);
cl::Buffer buf_b2(context, data_b.cbegin(), data_b.cend(), true);
cl::Buffer buf_b3(queue, data_c.cbegin(), data_c.cend(), true);
```

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -4002,7 +4002,7 @@ public:
         Context context = Context::getDefault(err);
 
         if( useHostPtr ) {
-            object_ = ::clCreateBuffer(context(), flags, size, static_cast<DataType*>(&*startIterator), &error);
+            object_ = ::clCreateBuffer(context(), flags, size, const_cast<DataType*>(&*startIterator), &error);
         } else {
             object_ = ::clCreateBuffer(context(), flags, size, 0, &error);
         }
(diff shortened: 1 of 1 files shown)


bashbaug (approved):
This seems reasonable to me but it'd be great to have another reviewer.  We need to use `const_cast` here because `clCreateBuffer` takes a `void*` parameter and not a `const void*`.

When the iterator is a const iterator does it ever make sense for `readOnly` to be `false`?  Seems like the buffer must be read-only, otherwise it's possible to modify the memory via a kernel or via host API calls, which probably isn't intended.  If `readOnly` must be `true` in this case is it worth trying to set it automatically?

pkeir:
It reminds of a similar case in SYCL:

"If the type of the host data is not const but the pointer to host data is const, then the read-only restriction applies only on host and not on device accesses.
When the buffer is destroyed, the destructor will block until all work in queues on the buffer have completed."

In our case, where "the pointer to host data is const", we might still allow read and write access on the device.

From point 2.b [here](https://www.khronos.org/registry/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:buf-sync-rules).

bashbaug:
> In our case, where "the pointer to host data is const", we might still allow read and write access on the device.

Makes sense to me.  I'll leave this open a couple more days just in case @kpet or @jrprice (or anyone else) is able to review, but unless there are objections I'll merge before the holidays.  Thank you!

kpet (approved):
LGTM. The exact behaviour with `CL_MEM_USE_HOST_PTR` is implementation-defined so I don't think we could mandate `CL_MEM_READ_ONLY` when using a const iterator with existing constructors. For implementations that do use the application-provided memory as the datastore for the memory object, this change means we're breaking the "const promise". Maybe we should introduce additional overloads that guarantee that the application-owned memory never gets modified?

I also note that the three modified constructors duplicate a lot of code. It seems trivial to rewrite them in terms of each other. This would greatly aid maintainability (probably a separate change).

bashbaug:
This slipped off of my TODO list - sorry about that!  Merging.

PR #170, opened 2021-12-10
Allow qualifying the OpenCL library calls with a namespace

fricc33:
Allow qualifying the OpenCL library calls with a namespace (CL_WRAPPER_NS) and to use function pointers to access the dynamic library (CL_USE_FUNCTION_POINTERS)

Dear author of opencl.hpp, I have a proposal for a change that would allow for easier integration with embedded systems such as Android.

On Android the user is not able to directly link against the OpenCL libraries, rather the app will have to discover the dll at run time and bind to the library at initialization using function pointers defined in cl_icd.h.

This approach causes problems when trying to use opencl.hpp, which expects canonical names for the OpenCL calls (e.g.: `::clGetContextInfo`) as we will have to define a different name for the function pointers to avoid clashes.

My solution to this problem is to define a separate C++ namespace for the function pointers, and refer to that namespace within opencl.hpp (see attached icd wrapper for reference).

The namespace can be defined through a preprocessor macro (`CL_WRAPPER_NS`) which can be left undefined by default, leaving the present opencl.hpp behavior unmodified. Wherever OpenCL calls are defined within the opencl.hpp source they are prefixed with the namespace macro (e.g.: `CL_WRAPPER_NS::clGetContextInfo`).

Finally, in the `GetInfoFunctor0` and `GetInfoFunctor1` structs, opencl.hpp uses function pointers already, which now become pointer to pointer breaking the code. We need to explicitly dereference the first indirection, for that I added another preprocessor macro (`CL_USE_FUNCTION_POINTERS`) which needs to be defined by the caller:

```
template <typename Func, typename Arg0>
struct GetInfoFunctor0
{
    Func f_; const Arg0& arg0_;
    cl_int operator ()(
        cl_uint param, size_type size, void* value, size_type* size_ret)
#if defined(CL_USE_FUNCTION_POINTERS)
    { return (*f_)(arg0_, param, size, value, size_ret); }
#else
    { return f_(arg0_, param, size, value, size_ret); }
#endif
};
```

This approach has been working very well in my Android applications and I thought it might be of use for others.

Thank you very much for this very useful library, cheers,
 - Fabio

[gls_icd_wrapper.zip](https://github.com/KhronosGroup/OpenCL-CLHPP/files/7694907/gls_icd_wrapper.zip)

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -396,6 +396,10 @@
 #ifndef CL_HPP_
 #define CL_HPP_
 
+#ifndef CL_WRAPPER_NS
+#define CL_WRAPPER_NS
+#endif
+
 /* Handle deprecated preprocessor definitions. In each case, we only check for
  * the old name if the new name is not defined, so that user code can define
  * both and hence work with either version of the bindings.
@@ -1349,7 +1353,9 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     \
     F(cl_image_info, CL_IMAGE_ARRAY_SIZE, size_type) \
     F(cl_image_info, CL_IMAGE_NUM_MIP_LEVELS, cl_uint) \
-    F(cl_image_info, CL_IMAGE_NUM_SAMPLES, cl_uint)
+    F(cl_image_info, CL_IMAGE_NUM_SAMPLES, cl_uint) \
+    F(cl_device_info, CL_DEVICE_IMAGE_PITCH_ALIGNMENT, cl_uint) \
+    F(cl_device_info, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT, cl_uint)
 
 #define CL_HPP_PARAM_NAME_INFO_2_0_(F) \
     F(cl_device_info, CL_DEVICE_QUEUE_ON_HOST_PROPERTIES, cl_command_queue_properties) \
@@ -1365,8 +1371,6 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     F(cl_device_info, CL_DEVICE_PREFERRED_PLATFORM_ATOMIC_ALIGNMENT, cl_uint) \
     F(cl_device_info, CL_DEVICE_PREFERRED_GLOBAL_ATOMIC_ALIGNMENT, cl_uint) \
     F(cl_device_info, CL_DEVICE_PREFERRED_LOCAL_ATOMIC_ALIGNMENT, cl_uint) \
-    F(cl_device_info, CL_DEVICE_IMAGE_PITCH_ALIGNMENT, cl_uint) \
-    F(cl_device_info, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT, cl_uint) \
     F(cl_device_info, CL_DEVICE_MAX_READ_WRITE_IMAGE_ARGS, cl_uint ) \
     F(cl_device_info, CL_DEVICE_MAX_GLOBAL_VARIABLE_SIZE, size_type ) \
     F(cl_device_info, CL_DEVICE_GLOBAL_VARIABLE_PREFERRED_TOTAL_SIZE, size_type ) \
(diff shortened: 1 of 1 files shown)


include/CL/opencl.hpp
```diff
@@ -1641,7 +1645,11 @@ struct GetInfoFunctor0
     Func f_; const Arg0& arg0_;
     cl_int operator ()(
         cl_uint param, size_type size, void* value, size_type* size_ret)
+#if defined(CL_USE_FUNCTION_POINTERS)
+    { return (*f_)(arg0_, param, size, value, size_ret); }
+#else
     { return f_(arg0_, param, size, value, size_ret); }
+#endif
```
bashbaug, line 1652: I'm not quite following why this change is needed, but it'd be really great if we could avoid needing an #ifdef for this.  A couple of options to consider:

1. Will the `CL_USE_FUNCTION_POINTERS` path work in all cases?  This seemed to work for the few cases I've tried and see also: https://godbolt.org/z/fMfTedq8d
2. If not, can we do something with partial template specialization when `Func` is a pointer type?

bashbaug (commented):
Thank you very much for your contribution!  I've left a few comments and suggestions for improvement.

include/CL/opencl.hpp
```diff
@@ -1349,7 +1353,9 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     \
     F(cl_image_info, CL_IMAGE_ARRAY_SIZE, size_type) \
     F(cl_image_info, CL_IMAGE_NUM_MIP_LEVELS, cl_uint) \
-    F(cl_image_info, CL_IMAGE_NUM_SAMPLES, cl_uint)
+    F(cl_image_info, CL_IMAGE_NUM_SAMPLES, cl_uint) \
+    F(cl_device_info, CL_DEVICE_IMAGE_PITCH_ALIGNMENT, cl_uint) \
+    F(cl_device_info, CL_DEVICE_IMAGE_BASE_ADDRESS_ALIGNMENT, cl_uint)
```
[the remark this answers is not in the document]
fricc33: Oh, I see, maybe we can address this in a separate pull request

include/CL/opencl.hpp
```diff
@@ -396,6 +396,10 @@
 #ifndef CL_HPP_
 #define CL_HPP_
 
+#ifndef CL_WRAPPER_NS
+#define CL_WRAPPER_NS
```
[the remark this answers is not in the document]
fricc33: Definitively =)

include/CL/opencl.hpp
fricc33: Oh, I'm sure this can be done more elegantly, the reason why this is needed is that when using function pointers to access the OpenCL API f_ becomes a pointer to a function pointer instead of a simple function pointer, and therefore you need to dereference once to get the function call syntax to work.

fricc33: Oh, I see what you mean, sure, explicitly dereferencing the pointer does work all the time =)

bashbaug (approved):
Thanks, the latest version addressed all of my comments.  I have a slight preference for a more descriptive name for the new macro (currently `CL_HPP_WRAPPER_NS`), but I can live with it as-is and I'm not sure how much to bike-shed this anyhow.  LGTM!

I'd like a couple of others from the working group to add their review and people are starting to disappear for vacation, but we'll aim to get this merged early in January.

Thanks again!

fricc33:
Hi Ben,
thank you very much for the review =)
I was thinking about a more descriptive name, maybe CL_HPP_DLL_WRAPPER_NS,
or CL_HPP_DLL_PTR_NS?
Thanks again, cheers,
 - Fabio

bashbaug:
Thank you for your patience, and sorry for taking so long to get back to you.

We discussed this PR in our working group meeting and our recommended name for the new macro is `CL_HPP_OPENCL_API_NAMESPACE`.  You've been very patient so I'm happy to help with the renaming if this would be helpful, likely as a PR to your branch.  I see this PR needs a rebase now also.

Let me know how you'd like to proceed.  As soon as the renaming is complete we're all set to merge.  Thanks again!

bashbaug:
I've created a PR against your branch with the renaming in case it's helpful: https://github.com/Glass-Imaging/OpenCL-CLHPP/pull/1

I've also checked the rebase and it doesn't look too bad (the conflict is due to the recent static_cast -> const_cast changes).

PR #172, opened 2022-01-25
Fix Doxygen template variable

jrprice:
The CLCPP_SOURCE_DIR CMake variable was removed when the CMake build
was recently reworked, so use PROJECT_SOURCE_DIR instead.


kpet (approved):
LGTM

PR #191, opened 2023-02-14
Add semaphore support

### LINKED ISSUE #154: Add support for cl_khr_semaphore

Mikolaj-Filar-Mobica:
Co-authored-by: Lukasz Pawlik <lukasz.pawlik@mobica.com>

Closes #154

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -744,6 +744,9 @@ namespace cl {
     class Memory;
     class Buffer;
     class Pipe;
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+    class SemaphoreKhr;
+#endif
 
 #if defined(CL_HPP_ENABLE_EXCEPTIONS)
     /*! \brief Exception class 
@@ -939,6 +942,12 @@ static inline cl_int errHandler (cl_int err, const char * errStr = NULL)
 #define __SET_PROGRAM_SPECIALIZATION_CONSTANT_ERR   CL_HPP_ERR_STR_(clSetProgramSpecializationConstant)
 #endif
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+#define __GET_SEMAPHORE_KHR_INFO_ERR                    CL_HPP_ERR_STR_(clGetSemaphoreInfoKHR)
+#define __CREATE_SEMAPHORE_KHR_WITH_PROPERTIES_ERR      CL_HPP_ERR_STR_(clCreateSemaphoreWithPropertiesKHR)
+#define __ENQUEUE_WAIT_SEMAPHORE_KHR_ERR                CL_HPP_ERR_STR_(clEnqueueWaitSemaphoresKHR)
+#define __ENQUEUE_SIGNAL_SEMAPHORE_KHR_ERR              CL_HPP_ERR_STR_(clEnqueueSignalSemaphoresKHR)
+#endif // CL_HPP_TARGET_OPENCL_VERSION >= 300
 
 /**
  * CL 1.2 version that uses device fission.
@@ -987,6 +996,43 @@ static inline cl_int errHandler (cl_int err, const char * errStr = NULL)
 #endif // CL_HPP_USER_OVERRIDE_ERROR_STRINGS
 //! \endcond
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300 
+typedef CL_API_ENTRY cl_semaphore_khr(CL_API_CALL* PFN_clCreateSemaphoreWithPropertiesKHR)(
+    cl_context context,
+    const cl_semaphore_properties_khr* sema_props,
+    cl_int* errcode_ret);
+
+typedef CL_API_ENTRY cl_int(CL_API_CALL* PFN_clReleaseSemaphoreKHR)(cl_semaphore_khr sema_object);
+typedef CL_API_ENTRY cl_int(CL_API_CALL* PFN_clRetainSemaphoreKHR)(cl_semaphore_khr sema_object);
+
+typedef CL_API_ENTRY cl_int(CL_API_CALL* PFN_clEnqueueWaitSemaphoresKHR)(
+    cl_command_queue command_queue, cl_uint num_sema_objects,
+    const cl_semaphore_khr* sema_objects,
+    const cl_semaphore_payload_khr* sema_payload_list,
+    cl_uint num_events_in_wait_list, const cl_event* event_wait_list,
+    cl_event* event);
+
+typedef CL_API_ENTRY cl_int(CL_API_CALL* PFN_clEnqueueSignalSemaphoresKHR)(
+    cl_command_queue command_queue, cl_uint num_sema_objects,
+    const cl_semaphore_khr* sema_objects,
+    const cl_semaphore_payload_khr* sema_payload_list,
+    cl_uint num_events_in_wait_list, const cl_event* event_wait_list,
+    cl_event* event);
+
+typedef CL_API_ENTRY cl_int(CL_API_CALL* PFN_clGetSemaphoreInfoKHR)(
+    cl_semaphore_khr sema_object, 
+    cl_semaphore_info_khr param_name,
+    ::size_t param_value_size,
+    void* param_value,
+    ::size_t* param_value_size_ret);
+
+CL_HPP_DEFINE_STATIC_MEMBER_ PFN_clCreateSemaphoreWithPropertiesKHR pfn_clCreateSemaphoreWithPropertiesKHR  = nullptr;
+CL_HPP_DEFINE_STATIC_MEMBER_ PFN_clReleaseSemaphoreKHR              pfn_clReleaseSemaphoreKHR               = nullptr;
+CL_HPP_DEFINE_STATIC_MEMBER_ PFN_clRetainSemaphoreKHR               pfn_clRetainSemaphoreKHR                = nullptr;
+CL_HPP_DEFINE_STATIC_MEMBER_ PFN_clEnqueueWaitSemaphoresKHR         pfn_clEnqueueWaitSemaphoresKHR          = nullptr;
+CL_HPP_DEFINE_STATIC_MEMBER_ PFN_clEnqueueSignalSemaphoresKHR       pfn_clEnqueueSignalSemaphoresKHR        = nullptr;
+CL_HPP_DEFINE_STATIC_MEMBER_ PFN_clGetSemaphoreInfoKHR              pfn_clGetSemaphoreInfoKHR               = nullptr;
+#endif // CL_HPP_TARGET_OPENCL_VERSION >= 300
 
 namespace detail {
 
@@ -1426,6 +1472,7 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
 #define CL_HPP_PARAM_NAME_INFO_3_0_(F) \
     F(cl_platform_info, CL_PLATFORM_NUMERIC_VERSION, cl_version) \
     F(cl_platform_info, CL_PLATFORM_EXTENSIONS_WITH_VERSION, cl::vector<cl_name_version>) \
+    F(cl_platform_info, CL_PLATFORM_SEMAPHORE_TYPES_KHR, cl::vector<cl_semaphore_type_khr>) \
     \
     F(cl_device_info, CL_DEVICE_NUMERIC_VERSION, cl_version) \
     F(cl_device_info, CL_DEVICE_EXTENSIONS_WITH_VERSION, cl::vector<cl_name_version>) \
@@ -1442,11 +1489,13 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     F(cl_device_info, CL_DEVICE_DEVICE_ENQUEUE_CAPABILITIES, cl_device_device_enqueue_capabilities) \
     F(cl_device_info, CL_DEVICE_PIPE_SUPPORT, cl_bool) \
     F(cl_device_info, CL_DEVICE_LATEST_CONFORMANCE_VERSION_PASSED, string) \
+    F(cl_device_info, CL_DEVICE_SEMAPHORE_TYPES_KHR, cl::vector<cl_semaphore_type_khr>) \
     \
     F(cl_command_queue_info, CL_QUEUE_PROPERTIES_ARRAY, cl::vector<cl_queue_properties>) \
     F(cl_mem_info, CL_MEM_PROPERTIES, cl::vector<cl_mem_properties>) \
     F(cl_pipe_info, CL_PIPE_PROPERTIES, cl::vector<cl_pipe_properties>) \
-    F(cl_sampler_info, CL_SAMPLER_PROPERTIES, cl::vector<cl_sampler_properties>)
+    F(cl_sampler_info, CL_SAMPLER_PROPERTIES, cl::vector<cl_sampler_properties>) \
+    F(cl_semaphore_info_khr, CL_SEMAPHORE_PROPERTIES_KHR, cl::vector<cl_semaphore_properties_khr>)
 
 template <typename enum_type, cl_int Name>
 struct param_traits {};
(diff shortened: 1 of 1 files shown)


include/CL/opencl.hpp
```diff
@@ -8859,6 +8931,54 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
     {
         return detail::errHandler(::clFinish(object_), __FINISH_ERR);
     }
+
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+    cl_int enqueueWaitSemaphoresKHR(
+        const vector<SemaphoreKhr>* sema_objects,
+        const vector<cl_semaphore_payload_khr>* sema_payloads = NULL,
```
bashbaug, line 8938: Could we do better than a pointer-to-vector here too?  Maybe:

```suggestion
        const vector<cl_semaphore_payload_khr>& sema_payloads = {},
```

include/CL/opencl.hpp
```diff
@@ -8859,6 +8931,54 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
...
+                (event != NULL) ? &tmp : NULL),
+            __ENQUEUE_WAIT_SEMAPHORE_KHR_ERR);
+        if (event != NULL && err == CL_SUCCESS)
+            *event = tmp;
+        return err;
+    }
+
+    cl_int enqueueSignalSemaphoreKhr(
+        const vector<SemaphoreKhr>* sema_objects,
+        const vector<cl_semaphore_payload_khr>* sema_payloads = NULL,
```
bashbaug, line 8960: Same comments as for `enqueueWaitSemaphoresKHR`:

```suggestion
        const vector<SemaphoreKhr>& sema_objects,
        const vector<cl_semaphore_payload_khr>& sema_payloads = {},
```

include/CL/opencl.hpp
```diff
@@ -10253,7 +10373,102 @@ namespace compatibility {
     };
 } // namespace compatibility
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+class SemaphoreKhr : public detail::Wrapper<cl_semaphore_khr>
```
bashbaug, line 10377: Bikeshedding: Should this be `SempahoreKhr` or `SemaphoreKHR` (or maybe even just `Semaphore`)?

I think I'd prefer just `Semaphore` personally...

bashbaug (changes requested):
This is a great start!

include/CL/opencl.hpp
```diff
@@ -744,6 +744,9 @@ namespace cl {
     class Memory;
     class Buffer;
     class Pipe;
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -1806,6 +1855,29 @@ struct ReferenceHandler<cl_event>
...
+template <>
+struct ReferenceHandler<cl_semaphore_khr>
+{
+    static cl_int retain(cl_semaphore_khr semaphore)
+    { 
+        if (pfn_clRetainSemaphoreKHR != nullptr) {
+            return pfn_clRetainSemaphoreKHR(semaphore);
+        }
+
+        return CL_INVALID_VALUE;
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -1426,6 +1472,7 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
 #define CL_HPP_PARAM_NAME_INFO_3_0_(F) \
     F(cl_platform_info, CL_PLATFORM_NUMERIC_VERSION, cl_version) \
     F(cl_platform_info, CL_PLATFORM_EXTENSIONS_WITH_VERSION, cl::vector<cl_name_version>) \
+    F(cl_platform_info, CL_PLATFORM_SEMAPHORE_TYPES_KHR, cl::vector<cl_semaphore_type_khr>) \
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -987,6 +996,43 @@ static inline cl_int errHandler (cl_int err, const char * errStr = NULL)
 #endif // CL_HPP_USER_OVERRIDE_ERROR_STRINGS
 //! \endcond
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300 
+typedef CL_API_ENTRY cl_semaphore_khr(CL_API_CALL* PFN_clCreateSemaphoreWithPropertiesKHR)(
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -8859,6 +8931,54 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
...
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+    cl_int enqueueWaitSemaphoresKHR(
+        const vector<SemaphoreKhr>* sema_objects,
+        const vector<cl_semaphore_payload_khr>* sema_payloads = NULL,
+        const vector<Event>* events_wait_list = NULL,
+        Event *event = NULL) const
+    {
+        cl_event tmp;
+        cl_int err = detail::errHandler(
+            pfn_clEnqueueWaitSemaphoresKHR(
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -10253,7 +10373,102 @@ namespace compatibility {
...
 } // namespace compatibility
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+class SemaphoreKhr : public detail::Wrapper<cl_semaphore_khr>
+{
+public:
+    SemaphoreKhr() : detail::Wrapper<cl_type>() {}
+    SemaphoreKhr(
+        const cl::Device& device,
+        const cl_semaphore_properties_khr *sema_props,
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -8859,6 +8931,54 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
     {
         return detail::errHandler(::clFinish(object_), __FINISH_ERR);
     }
+
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+    cl_int enqueueWaitSemaphoresKHR(
+        const vector<SemaphoreKhr>* sema_objects,
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Done, thank you! I've used `Semaphore`.

MathiasMagnus (changes requested):
I think the CTOR interface to semaphores is a bit lacking and likely buggy without changes.

include/CL/opencl.hpp
```diff
@@ -10253,7 +10371,103 @@ namespace compatibility {
...
+{
+public:
+    Semaphore() : detail::Wrapper<cl_type>() {}
+    Semaphore(
+        const cl::Device& device,
+        const vector<cl_semaphore_properties_khr>& sema_props,
+        cl_int *err = NULL) 
+    {
+        cl_int error;
+        Context context = Context::getDefault(&error);
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: @MathiasMagnus, do you prefer other constructor or additional parameter in existing one (`Context *context`)?

include/CL/opencl.hpp
```diff
@@ -10253,7 +10373,102 @@ namespace compatibility {
...
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+class SemaphoreKhr : public detail::Wrapper<cl_semaphore_khr>
+{
+public:
+    SemaphoreKhr() : detail::Wrapper<cl_type>() {}
+    SemaphoreKhr(
+        const cl::Device& device,
+        const cl_semaphore_properties_khr *sema_props,
+        cl_int *err = NULL)
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
```diff
@@ -10253,7 +10373,102 @@ namespace compatibility {
...
+    SemaphoreKhr(
+        const cl::Device& device,
+        const cl_semaphore_properties_khr *sema_props,
+        cl_int *err = NULL) 
+    {
+        cl_int error;
+        Context context = Context::getDefault(&error);
+ 
+        if (error == CL_SUCCESS) {
+            error = initExtensions(device);
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: Done, thank you! I've added the new constructor.

include/CL/opencl.hpp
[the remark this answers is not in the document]
bashbaug: This is mostly done - thanks!  I think there are a few additional NULL checks that would be good to add though, and I'll add specific review comments for them.

include/CL/opencl.hpp
```diff
@@ -10253,7 +10370,118 @@ namespace compatibility {
...
+        return *this;
+    }
+    template <typename T>
+    cl_int getInfo(cl_semaphore_info_khr name, T* param) const
+    {
+        typedef CL_API_ENTRY cl_int(CL_API_CALL * PFN_clGetSemaphoreInfoKHR)(
+            cl_semaphore_khr sema_object, cl_semaphore_info_khr param_name,
+            size_type param_value_size, void* param_value,
+            size_type* param_value_size_ret);
+
```
bashbaug, line 10416: I think this is leftover code and can be removed?

include/CL/opencl.hpp
```diff
@@ -10253,7 +10370,118 @@ namespace compatibility {
...
+        return detail::errHandler(
+            detail::getInfo(&pfn_clGetSemaphoreInfoKHR, object_, name, param),
```
bashbaug, line 10418: It'd be a little weird but I think we can still get a possible `nullptr` function pointer call here by default constructing a semaphore and then calling `getInfo` on it, so I'd recommend adding a `nullptr` funciton pointer check.

include/CL/opencl.hpp
```diff
@@ -10253,7 +10370,118 @@ namespace compatibility {
...
+
+    cl_int retain()
+    { 
+        return pfn_clRetainSemaphoreKHR(object_);
+    }
+
+    cl_int release()
+    { 
+        return pfn_clReleaseSemaphoreKHR(object_);
+    }
```
bashbaug, line 10442: Same goes here - it'd be a little strange, but we could call `nullptr` function pointers here by calling `retain` or `release` on a default constructed semaphore.  It'd be best to add `nullptr` function pointer checks.

include/CL/opencl.hpp
```diff
@@ -10253,7 +10370,118 @@ namespace compatibility {
...
+            (pfn_clGetSemaphoreInfoKHR              == nullptr))
+        {
+            ext_init_error_ = CL_INVALID_VALUE;
+        }
+    }
+
+};
+
+CL_HPP_DEFINE_STATIC_MEMBER_ std::once_flag Semaphore::ext_init_;
+CL_HPP_DEFINE_STATIC_MEMBER_ cl_int Semaphore::ext_init_error_ = CL_SUCCESS;
```
bashbaug, line 10482: I think I understand why this static error variable exists (it ensures a semaphore cannot be created unless all function pointers are properly queried) and I'm OK having it, but an alternative we could consider is to allow `nullptr` function pointers and only return an error if/when trying to use a `nullptr` function pointer.  What do you think?

bashbaug (approved):
Thanks!  I'm fine with this going in as-is, but here are a few final nice-to-haves.

I'll try to port my semaphore sample to use these changes just to see if there are any issues in practice.

Have you considered adding any semaphore unit tests?

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: The only purpose for these variables is to satisfy the call_once for `InitExtensions`. `initExtensions` is called in the constructor and one of the pointers (`pfn_clCreateSemaphoreWithPropertiesKHR`) is called after that, so we would need to throw the exception there anyway or shall we skip the function call? 
I'm wondering if there could be a scenario where someone would want to call it for multiple devices (e.g. in a loop trying to find one that supports semaphores).

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Indeed, code removed. thank you!

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Ahh, I hadn't thought about the default constructor. Done, thank you!

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Done, thank you!

include/CL/opencl.hpp
bashbaug: Good questions, I think this discussion really amounts to: how do we want to behave if one or more semaphore function pointers cannot be queried?

Option 1: `initExtensions` succeeds but sets a static variable to an error condition.  Any attempts to create a semaphore fail.  This is the current behavior.

Option 2: `initExtensions` succeeds, but any attempts to use a `nullptr` function pointer will fail.  This might allow some semaphore functionality, depending on which function pointers are `nullptr`.  This was my alternative suggested above.

Option 3: `initExtensions` fails (throws an exception) and the attempt to create a semaphore fails.  Trying to create a semaphore with a different context with a different device may succeed.  This is a little bit like Option 1, but with the retry behavior.

Note that if we want to go with Option 3 we will need to throw an exception from `initExtension`, otherwise `std::call_once` will not try again:  https://en.cppreference.com/w/cpp/thread/call_once

Mikolaj-Filar-Mobica: Thanks for the note. I wasn't aware of that :)
Regarding option 2, is it possible that not all pointers are correctly initialized for some platforms? I don't know all the usage scenarios, but options 2 and 3 look fine to me. Which one do you prefer?

include/CL/opencl.hpp
```diff
@@ -8859,6 +8920,64 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
     {
         return detail::errHandler(::clFinish(object_), __FINISH_ERR);
     }
+
+#ifdef cl_khr_semaphore
+    cl_int enqueueWaitSemaphores(
+        const vector<Semaphore> &sema_objects,
+        const vector<cl_semaphore_payload_khr> &sema_payloads = {},
+        const vector<Event>* events_wait_list = NULL,
+        Event *event = NULL) const
```
keryell, line 8929: In new code, do not use macro `NULL`.
```suggestion
        const vector<Event>* events_wait_list = nullptr,
        Event *event = nullptr) const
```

Mikolaj-Filar-Mobica: Shouldn't we change it in the whole file? It looks inconsistent now: there are more than 600 occurrences of the `NULL` and about 40 of the `nullptr`. If so, do you think it's a good idea to do it in another task?

keryell: Yes probably a good task. The inconsistency is not my main concern when I look at the file as an antique shop... :-)

Mikolaj-Filar-Mobica: Done

include/CL/opencl.hpp
bashbaug: Let's go with Option 2 for now.  I think this is most consistent with the extension function pointer usage elsewhere in this file.  I'll open a separate issue to revisit extension handling in general, but we can solve that with separate PRs.

So, specific tasks:
* Remove all of the `nullptr` checks from `initExtensions`.
* Remove `ext_init_error_`.
* Add a check for a `nullptr` `pfn_clCreateSemaphoreWithPropertiesKHR` in the `Semaphore` constructor, since we will no longer be checking against `ext_init_error_`.

Sound good?

For "bonus points" we could throw an exception and retry if _none_ of the extension function pointers are non-`nullptr`, but this is probably overkill.

bashbaug: Thanks!  Latest update looks great 👍

MathiasMagnus (approved):
Good to go.

PR #192, opened 2023-02-14
Create cpp wrapper for cl_command_buffer_khr

### LINKED ISSUE #188: Add support for cl_command_buffer_khr
Add C++ wrapper for cl_command_buffer_khr (https://github.com/KhronosGroup/OpenCL-Headers/blob/main/CL/cl_ext.h).

Lukasz-Krakowiak-Mobica:
Closes  #188

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -745,6 +745,9 @@ namespace cl {
 #ifdef cl_khr_semaphore
     class Semaphore;
 #endif
+#if defined(cl_khr_command_buffer)
+    class CommandBufferKhr;
+#endif // cl_khr_command_buffer
 
 #if defined(CL_HPP_ENABLE_EXCEPTIONS)
     /*! \brief Exception class 
@@ -951,6 +954,21 @@ static inline cl_int errHandler (cl_int err, const char * errStr = nullptr)
 #define __RETAIN_SEMAPHORE_KHR_ERR                      CL_HPP_ERR_STR_(clRetainSemaphoreKHR)
 #define __RELEASE_SEMAPHORE_KHR_ERR                     CL_HPP_ERR_STR_(clReleaseSemaphoreKHR)
 #endif
+#if defined(cl_khr_command_buffer)
+#define __CREATE_COMMAND_BUFFER_KHR_ERR             CL_HPP_ERR_STR_(clCreateCommandBufferKHR)
+#define __GET_COMMAND_BUFFER_INFO_KHR_ERR           CL_HPP_ERR_STR_(clGetCommandBufferInfoKHR)
+#define __FINALIZE_COMMAND_BUFFER_KHR_ERR           CL_HPP_ERR_STR_(clFinalizeCommandBufferKHR)
+#define __ENQUEUE_COMMAND_BUFFER_KHR_ERR            CL_HPP_ERR_STR_(clEnqueueCommandBufferKHR)
+#define __COMMAND_BARRIER_WITH_WAIT_LIST_KHR_ERR    CL_HPP_ERR_STR_(clCommandBarrierWithWaitListKHR)
+#define __COMMAND_COPY_BUFFER_KHR_ERR               CL_HPP_ERR_STR_(clCommandCopyBufferKHR)
+#define __COMMAND_COPY_BUFFER_RECT_KHR_ERR          CL_HPP_ERR_STR_(clCommandCopyBufferRectKHR)
+#define __COMMAND_COPY_BUFFER_TO_IMAGE_KHR_ERR      CL_HPP_ERR_STR_(clCommandCopyBufferToImageKHR)
+#define __COMMAND_COPY_IMAGE_KHR_ERR                CL_HPP_ERR_STR_(clCommandCopyImageKHR)
+#define __COMMAND_COPY_IMAGE_TO_BUFFER_KHR_ERR      CL_HPP_ERR_STR_(clCommandCopyImageToBufferKHR)
+#define __COMMAND_FILL_BUFFER_KHR_ERR               CL_HPP_ERR_STR_(clCommandFillBufferKHR)
+#define __COMMAND_FILL_IMAGE_KHR_ERR                CL_HPP_ERR_STR_(clCommandFillImageKHR)
+#define __COMMAND_NDRANGE_KERNEL_KHR_ERR            CL_HPP_ERR_STR_(clCommandNDRangeKernelKHR)
+#endif // cl_khr_command_buffer
 
 /**
  * CL 1.2 version that uses device fission.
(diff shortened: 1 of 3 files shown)


EwanC (commented):
I've never used or developed in this repo before, so can't review in-depth and only have one small comment. Looks fine to me otherwise.

include/CL/opencl.hpp
```diff
@@ -988,6 +1006,167 @@ static inline cl_int errHandler (cl_int err, const char * errStr = NULL)
 //! \endcond
 
 
+#if defined(cl_khr_command_buffer)
```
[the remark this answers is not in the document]
Mikolaj-Filar-Mobica: I see the `cl_ext.h` file is included in `opencl.hpp` conditionally:
```C++
#if defined(CL_HPP_USE_CL_DEVICE_FISSION) || defined(CL_HPP_USE_CL_SUB_GROUPS_KHR)
#include <CL/cl_ext.h>
#endif
```
I don't know these definitions. Is the inclusion guaranteed?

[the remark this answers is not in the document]
bashbaug: Hrm, this seems like a bug.  Because we're including `CL/opencl.h` unconditionally, which includes `CL/cl_ext.h` unconditionally, there's no reason to conditionally include `CL/cl_ext.h`.  I'll file a PR to fix this.

This also means that EwanC's comment is correct and we can use the function pointer typedefs in `cl_ext.h` vs. duplicating them here.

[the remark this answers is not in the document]
bashbaug: PR to fix (remove) the conditional inclusion is here: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/195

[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: Thanks for you comments.

This redeclaration function pointers is requirement by macro: https://github.com/KhronosGroup/OpenCL-CLHPP/blob/main/include/CL/opencl.hpp#L733. It is adding prefix 'PFN_'. I used already existing impl but I am agree, it would be better to reuse the ones defined in cl_ext.h.

Could you advise me the best solution for this, shall I create a new Macro ?

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+        typename detail::param_traits<
+            detail::cl_command_buffer_info_khr, name>::param_type param;
+        cl_int result = getInfo(name, &param);
+        if (err != NULL) {
+            *err = result;
+        }
+        return param;
+    }
+
+    cl_int finalizeCommandBuffer()
```
bananAshkar, line 10508: should the methods be const , what I mean is should we replace "cl_int finalizeCommandBuffer()" to "cl_int finalizeCommandBuffer() const"?

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+                mutable_handle),
+            __COMMAND_FILL_BUFFER_KHR_ERR);
+
+        if (sync_point != NULL && error == CL_SUCCESS)
+            *sync_point = tmp_sync_point;
+
+        return error;
+    }
+
+    cl_int commandFillImage(const CommandQueue* command_queue,
```
bananAshkar, line 10742: command_queue could be NULL,so why we can not put it optional parameter?

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Yes you are right, it should be const.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: I will fix this for all methods where command_queue could be NULL.

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+                (sync_point != NULL) ? &tmp_sync_point : NULL,
+                mutable_handle),
+            __COMMAND_NDRANGE_KERNEL_KHR_ERR);
+
+        if (sync_point != NULL && error == CL_SUCCESS)
+            *sync_point = tmp_sync_point;
+
+        return error;
+    }
+
```
bananAshkar, line 10802: there are two more functions that are related to command buffer (mutable dispatch) that we can support in c++ binding : 
clUpdateMutableCommandsKHR() , clGetMutableCommandInfoKHR()

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+        detail::Wrapper<cl_type>(commandBufferKhr, retainObject) { }
+
+    CommandBufferKhr& operator = (const cl_command_buffer_khr& rhs)
+    {
+        detail::Wrapper<cl_type>::operator=(rhs);
+        return *this;
+    }
+
+    /*! \brief Copy constructor to forward copy to the superclass correctly.
+     * Required for MSVC.
```
keryell, line 10460: Is this still true?

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+     */
+    CommandBufferKhr(const CommandBufferKhr& commandBufferKhr) : detail::Wrapper<cl_type>(commandBufferKhr) {}
```
keryell, line 10462: I have the feeling all these forwarding could be replaced by just
```c++
using CommandBufferKhr::CommandBufferKhr;
```

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+     */
+    CommandBufferKhr(CommandBufferKhr&& commandBufferKhr) CL_HPP_NOEXCEPT_ : detail::Wrapper<cl_type>(std::move(commandBufferKhr)) {}
+
+    /*! \brief Move assignment to forward move to the superclass correctly.
+     * Required for MSVC.
+     */
+    CommandBufferKhr& operator = (CommandBufferKhr &&commandBufferKhr)
+    {
+        detail::Wrapper<cl_type>::operator=(std::move(commandBufferKhr));
+        return *this;
```
keryell, line 10484: Not clear why all these are necessary and the one synthesized by the compiler are not good.

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+        const vector<cl_sync_point_khr>* sync_points_vec = NULL,
+        cl_sync_point_khr* sync_point = NULL,
+        cl_mutable_command_khr* mutable_handle = NULL)
+    {
+        cl_sync_point_khr tmp_sync_point;
+        cl_int error = detail::errHandler(
+            pfn_clCommandBarrierWithWaitListKHR(object_,
+                (command_queue != NULL) ? command_queue->operator()() : NULL,
+                (sync_points_vec != NULL) ? (cl_uint) sync_points_vec->size() : 0,
+                (sync_points_vec != NULL && sync_points_vec->size() > 0) ? (cl_sync_point_khr*) &sync_points_vec->front() : NULL,
```
keryell, line 10537: Why the cast?

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+        size_type dst_offset,
+        size_type size,
+        const vector<cl_sync_point_khr>* sync_points_vec = NULL,
+        cl_sync_point_khr* sync_point = NULL,
+        cl_mutable_command_khr* mutable_handle = NULL)
+    {
+        cl_sync_point_khr tmp_sync_point;
+        cl_int error = detail::errHandler(
+            pfn_clCommandCopyBufferKHR(object_,
+                (command_queue != NULL) ? command_queue->operator()() : NULL,
```
keryell, line 10561:
```suggestion
                (command_queue != NULL) ? command_queue->operator()() : NULL,
```
```suggestion
                (command_queue != NULL) ? (*command_queue)() : NULL,
```

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+#if defined(cl_khr_command_buffer)
+/*! \class CommandBufferKhr
+ * \brief CommandBufferKhr interface for cl_command_buffer_khr.
+ */
+class CommandBufferKhr : public detail::Wrapper<cl_command_buffer_khr>
+{
+private:
+    static std::once_flag default_initialized_;
+    static CommandBufferKhr default_;
+    static cl_int default_error_;
```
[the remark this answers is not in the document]
bananAshkar: I think that the use of std::once_flag suggests that the code is thread-safe , because these static variables happens only once..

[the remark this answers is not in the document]
keryell: This is not what I was thinking to but more what is the goal of this static variables at the first place? What happened if 2 command buffers are used at the same time in different threads
Please add a Doxygen-like comment on each of them explaining what they are for, otherwise the reader might think they are useless. ;-)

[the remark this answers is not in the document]
bashbaug: I actually don't think it makes much sense to have a "default" command buffer like it does to have a default context or command queue, and I'd suggest removing all of the "default" command buffer handling.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: OK Ben. For clGetMutableCommandInfoKHR best will be to create separately class MutableCommandKhr. Do you think is a proper way ?

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+            *errcode_ret = error;
+        }
+    }
+
+    explicit CommandBufferKhr(const cl_command_buffer_khr& commandBufferKhr, bool retainObject = false) :
+        detail::Wrapper<cl_type>(commandBufferKhr, retainObject) { }
+
+    CommandBufferKhr& operator = (const cl_command_buffer_khr& rhs)
+    {
+        detail::Wrapper<cl_type>::operator=(rhs);
```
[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: I modified it manually. I used same formatting as for other classes.

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+    cl_int getInfo(cl_command_buffer_info_khr name, T* param) const
+    {
+        return detail::errHandler(
+            detail::getInfo(pfn_clGetCommandBufferInfoKHR, object_, name, param),
+                __GET_COMMAND_BUFFER_INFO_KHR_ERR);
+    }
+
+    template <cl_command_buffer_info_khr name> typename
+        detail::param_traits<detail::cl_command_buffer_info_khr, name>::param_type
+        getInfo(cl_int* err = NULL) const
```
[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: OK

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: OK

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: No needed.

include/CL/opencl.hpp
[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: OK

include/CL/opencl.hpp
```diff
@@ -10253,6 +10342,530 @@ namespace compatibility {
...
+        detail::errHandler(default_error_, __CREATE_COMMAND_BUFFER_KHR_ERR);
+        if (err != NULL) {
+            *err = default_error_;
+        }
+        return default_;
+    }
+
+    static CommandBufferKhr setDefault(const CommandBufferKhr &default_cmd_buffer_khr)
+    {
+        std::call_once(default_initialized_, makeDefaultProvided, std::cref(default_cmd_buffer_khr));
```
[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: No needed any more.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: I wasn't sure of that. I am working on Linux OS and I had to ask colleague to verify this on Windows. All tests passed, so I removed this part of code.

include/CL/opencl.hpp
bananAshkar: I think that no need for class for MutableCommandKhr , you can just implement this like all the functions in commandBuffer class right?

include/CL/opencl.hpp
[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: I used c++ aliases to _fn pointers from cl_ext.h.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Done, thanks.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Done, thanks.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Done, thanks.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: I think it would be better to have this as separated class, but maybe I am wrong.
Please check my following commit.

include/CL/opencl.hpp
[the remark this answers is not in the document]
Lukasz-Krakowiak-Mobica: Done, thanks.

tests/test_openclhpp.cpp
```diff
@@ -3035,5 +3068,99 @@ void testDevicePCIBusInfo_KHR()
...
+// Stub for clGetCommandBufferInfoKHR that returns 1
+static cl_int clGetCommandBufferInfoKHR_testCommandBufferKhrGetNumQueues(
+    cl_command_buffer_khr command_buffer,
+    cl_command_buffer_info_khr param_name,
+    size_t param_value_size,
+    void *param_value,
+    size_t *param_value_size_ret,
+    int num_calls)
+{
+    (void) num_calls;
```
keryell, line 3096: If you do not use some parameters, do not name them
```suggestion
    int /* num_calls */)
{
```

include/CL/opencl.hpp
```diff
@@ -10253,6 +10369,482 @@ namespace compatibility {
...
+    explicit CommandBufferKhr(const vector<CommandQueue> &queues,
+        cl_command_buffer_properties_khr properties = 0,
+        cl_int* errcode_ret = nullptr)
+    {
+        cl_command_buffer_properties_khr command_buffer_properties[] = {
+            CL_COMMAND_BUFFER_FLAGS_KHR, properties, 0
+        };
+
+        /* initialization of addresses to extension functions (it is done only once) */
+        std::call_once(ext_init_, initExtensions, queues[0].getInfo<CL_QUEUE_DEVICE>());
```
keryell, line 10391: No pre-C++11 code

```suggestion
        std::call_once(ext_init_, [&] { initExtensions(queues[0].getInfo<CL_QUEUE_DEVICE>()); });
```

include/CL/opencl.hpp
```diff
@@ -10253,6 +10369,482 @@ namespace compatibility {
...
+    MutableCommandKhr() : detail::Wrapper<cl_type>() { }
+
+    explicit MutableCommandKhr(const cl_mutable_command_khr& mutableCommandKhr, bool retainObject = false) :
+        detail::Wrapper<cl_type>(mutableCommandKhr, retainObject) { }
+
+    MutableCommandKhr& operator=(const cl_mutable_command_khr& rhs)
+    {
+        detail::Wrapper<cl_type>::operator=(rhs);
+        return *this;
+    }
```
keryell, line 10824: Why do you need such an operator?

include/CL/opencl.hpp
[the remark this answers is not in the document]
keryell: But more fundamentally, why do you need such an operator?

keryell (commented):
A few nitpicks.

tests/test_openclhpp.cpp
```diff
@@ -3035,5 +3068,99 @@ void testDevicePCIBusInfo_KHR()
...
+    clRetainCommandQueue_ExpectAndReturn(make_command_queue(1), CL_SUCCESS);
+    clRetainCommandQueue_ExpectAndReturn(make_command_queue(2), CL_SUCCESS);
+
+    VECTOR_CLASS<cl::CommandQueue> command_queues = commandBufferKhrPool[0].getInfo<CL_COMMAND_BUFFER_QUEUES_KHR>();
+    TEST_ASSERT_EQUAL(3, command_queues.size());
+    TEST_ASSERT_EQUAL_PTR(make_command_queue(0), command_queues[0]());
+    TEST_ASSERT_EQUAL_PTR(make_command_queue(1), command_queues[1]());
+    TEST_ASSERT_EQUAL_PTR(make_command_queue(2), command_queues[2]());
+
+    command_queues[0]() = nullptr;
```
keryell, line 3160: @bashbaug when I look at this obfuscated code I really think we lack this interface in the `cl::Wrapper`
```suggestion
    command_queues[0].set(nullptr);
```
@Lukasz-Krakowiak-Mobica This is just for exposition.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: OK. Done

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: For any scenario where object type cl_mutable_command_khr will be assigned to object class MutableCommandKhr.
Do you think is  this unreal case ?

tests/test_openclhpp.cpp
Lukasz-Krakowiak-Mobica: Sure, done.

tests/test_openclhpp.cpp
Lukasz-Krakowiak-Mobica: I am trying to have consistence impl with rest of the the source. Do you really want me to change this ?

include/CL/opencl.hpp
keryell: No, it is. But why the current code does not get one synthesized automatically?
You have already a `MutableCommandKhr` constructor from a `const cl_mutable_command_khr&`, so I guess the default `MutableCommandKhr::operator=()` would kick in.
Just some naive question since I do not have the big picture.

Lukasz-Krakowiak-Mobica: No, it doesn't. This copy constructor will do this job if it won't have explicit specifier.

tests/test_openclhpp.cpp
bashbaug: I'm OK adding a `set()` function in the wrapper if we want one, but we could also just write:

```c++
command_queues[0] = nullptr;
```

Unless I'm missing something there's no reason to call `operator()`.... and actually it is almost certainly a bug to do so because it won't properly release reference counts on the object.

keryell: Ah good point. Then perhaps `operator()` should be `const` to avoid such bugs.

include/CL/opencl.hpp
keryell: Then perhaps this is a case where an implicit constructor makes sense instead of adding a lot of operators trying to mimic it?

tests/test_openclhpp.cpp
Lukasz-Krakowiak-Mobica: I think it isn't a bug. This is UT, and it helps only to prevent executing release() method: https://github.com/KhronosGroup/OpenCL-CLHPP/blob/main/include/CL/opencl.hpp#L1881 and doesn't change clReleaseCommandQueue_refcount.
But still if you will have any related changes I can apply them.

keryell: This code looks super convoluted, has no comment and a few people does not understand it.
Why do you want to circumvent the normal reference counting behavior?
We love comments in the code.
Think that ChatGPT will use your tests as a reference implementation to be used everywhere. :-)

bashbaug: Note, this is a pretty common pattern we use elsewhere in these unit tests even if it's not best practice or something we'd expect to see in real code.  An additional comment wouldn't hurt but I don't think we should refactor the tests to avoid this code, or at least not in this PR.

Example from a different test:

```c++
    // Prevent destructor from interfering with the test
    d0() = NULL;
```

Lukasz-Krakowiak-Mobica: Sorry guys, I used similar impl as it was done for other UTs.
I applied some changes for this UT, it should be more clear now.

keryell (approved):
Thanks.

Mikolaj-Filar-Mobica:
> Could we be sure this PR isn't affected by the same issue described in #206?
> 
> Until we get CI with this configuration we'll probably need to test this manually. I have some CI changes prototyped here, but this is admittedly a hack and will need modifications to become production-worthy:
> 
> https://github.com/bashbaug/OpenCL-CLHPP/tree/add-bleeding-edge-config

I was able to reproduce this problem with Semaphore on Windows by changing the compiler to _Intel C++ Compiler 2023_ (LLVM based) and setting the C++ language standard to _ISO C++20_ or _Preview_.
**I don't see a problem with this branch (merged with main)**
image

Lukasz-Krakowiak-Mobica:
> Could we be sure this PR isn't affected by the same issue described in #206?
> 
> Until we get CI with this configuration we'll probably need to test this manually. I have some CI changes prototyped here, but this is admittedly a hack and will need modifications to become production-worthy:
> 
> https://github.com/bashbaug/OpenCL-CLHPP/tree/add-bleeding-edge-config

Hi Ben,
I don't have here a case where I am using vector with any incomplete type.
I also compiled this branch with g++-8 and change in CMakeList.txt: set(CMAKE_CXX_STANDARD 20). Everything passed with success.

I am not sure how I can use .yml files from your branch.

bashbaug:
I rebased my "bleeding edge config" onto this PR and I didn't see any issues so I think we're good to go.

https://github.com/bashbaug/OpenCL-CLHPP/actions/runs/4548389646/jobs/8019372195

Thanks!  Merging as discussed in the March 28th teleconference.

PR #194, opened 2023-02-15
Add pkg-config file for C++ API bindings

Kerilk:
This implements pkg-config support for C++ bindings, which would be good to have, as hinted by https://github.com/KhronosGroup/OpenCL-Headers/issues/217

I will add a commit updating the path of the regular headers to share, but this will break the CI until https://github.com/KhronosGroup/OpenCL-Headers/pull/218 is merged.

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -2,6 +2,9 @@ name: Linux
 
 on: [push, pull_request]
 
+env:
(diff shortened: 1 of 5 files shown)


bashbaug (approved):
Usual caveat: I am not a pkg-config expert.  This matches the similar changes in OpenCL-Headers though and CI tests are passing, so LGTM!

PR #200, opened 2023-03-09
replace NULL with nullptr in opencl.hpp file

pwisniewskimobica:
All the NULL keywords were changed to nullptr

### DIFF
--- a/tests/test_openclhpp.cpp
+++ b/tests/test_openclhpp.cpp
@@ -86,10 +86,10 @@ static cl_int clGetCommandQueueInfo_context(
 
 
     TEST_ASSERT_EQUAL_HEX(CL_QUEUE_CONTEXT, param_name);
-    TEST_ASSERT(param_value == NULL || param_value_size >= sizeof(cl_context));
-    if (param_value_size_ret != NULL)
+    TEST_ASSERT(param_value == nullptr || param_value_size >= sizeof(cl_context));
+    if (param_value_size_ret != nullptr)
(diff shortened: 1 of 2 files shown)


tests/test_openclhpp.cpp
```diff
@@ -153,13 +153,13 @@ static cl_int clGetPlatformInfo_version(
 {
     size_t bytes = strlen(version) + 1;
 
-    TEST_ASSERT_NOT_NULL(id);
+    TEST_ASSERT_NOT_nullptr(id);
```
bashbaug, line 156: I think your search-and-replace was a little too aggressive :smiley:.

These should all remain `TEST_ASSERT_NOT_NULL`.

bashbaug (changes requested):
Mostly LGTM but found one issue that needs to be fixed.

Also needs a trivial rebase.

Thanks!

tests/test_openclhpp.cpp
pwisniewskimobica: Thanks for this!
Just fixed

keryell (approved):
Thanks for cleaning 12 years of antiques. :-)

bashbaug:
Looks like something went wrong in the rebase and there are CI failures now - please take a look.

pwisniewskimobica:
Rebased again. thanks

bashbaug (changes requested):

bashbaug (approved):
Thanks, all good now!

PR #201, opened 2023-03-09
Add constructor of NDRange for std::array argument and wrappers for i…

Mikolaj-Filar-Mobica:
…mage operations for 2D arguments

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -5935,6 +5935,15 @@ public:
         sizes_[2] = size2;
     }
 
+    //! \brief Constructs one-dimensional range.
+    NDRange(array<size_type, 1> a) : NDRange(a[0]){}
+
+    //! \brief Constructs two-dimensional range.
+    NDRange(array<size_type, 2> a) : NDRange(a[0], a[1]){}
+
+    //! \brief Constructs three-dimensional range.
+    NDRange(array<size_type, 3> a) : NDRange(a[0], a[1], a[2]){}
+
     /*! \brief Conversion operator to const size_type *.
      *  
      *  \returns a pointer to the size of the first dimension.
@@ -7858,6 +7867,35 @@ public:
         return err;
     }
 
+    cl_int enqueueReadBufferRect(
+        const Buffer& buffer,
+        cl_bool blocking,
+        const array<size_type, 2>& buffer_offset,
+        const array<size_type, 2>& host_offset,
+        const array<size_type, 2>& region,
+        size_type buffer_row_pitch,
+        size_type buffer_slice_pitch,
+        size_type host_row_pitch,
+        size_type host_slice_pitch,
+        void* ptr,
+        const vector<Event>* events = nullptr,
+        Event* event = nullptr) const
+    { 
+        return enqueueReadBufferRect(
+            buffer,
+            blocking,
+            { buffer_offset[0], buffer_offset[1], 0 },
+            { host_offset[0], host_offset[1], 0 },
+            { region[0], region[1], 1 },
+            buffer_row_pitch,
+            buffer_slice_pitch,
+            host_row_pitch,
+            host_slice_pitch,
+            ptr,
+            events,
+            event);
+    }
+
     cl_int enqueueWriteBufferRect(
         const Buffer& buffer,
         cl_bool blocking,
@@ -7897,6 +7935,35 @@ public:
         return err;
     }
 
+    cl_int enqueueWriteBufferRect(
+        const Buffer& buffer,
+        cl_bool blocking,
+        const array<size_type, 2>& buffer_offset,
+        const array<size_type, 2>& host_offset,
+        const array<size_type, 2>& region,
+        size_type buffer_row_pitch,
+        size_type buffer_slice_pitch,
+        size_type host_row_pitch,
+        size_type host_slice_pitch,
+        const void* ptr,
+        const vector<Event>* events = nullptr,
+        Event* event = nullptr) const
+    {
+        return enqueueWriteBufferRect(
+            buffer, 
+            blocking,
+            { buffer_offset[0], buffer_offset[1], 0 },
+            { host_offset[0], host_offset[1], 0 },
+            { region[0], region[1], 1 },
+            buffer_row_pitch,
+            buffer_slice_pitch,
+            host_row_pitch,
+            host_slice_pitch,
+            ptr,
+            events,
+            event);
+    }
+
     cl_int enqueueCopyBufferRect(
         const Buffer& src,
         const Buffer& dst,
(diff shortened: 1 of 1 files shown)


include/CL/opencl.hpp
```diff
@@ -5838,6 +5838,15 @@ class NDRange
...
     }
 
+    //! \brief Constructs one-dimensional range.
+    NDRange(std::array<size_type, 1> a) : NDRange(a[0]){}
+
+    //! \brief Constructs two-dimensional range.
+    NDRange(std::array<size_type, 2> a) : NDRange(a[0], a[1]){}
+
+    //! \brief Constructs three-dimensional range.
+    NDRange(std::array<size_type, 3> a) : NDRange(a[0], a[1], a[2]){}
```
bashbaug, line 5848: I think we should use the un-namespaced `array` class here rather than `std::array`, to work with `CL_HPP_NO_STD_ARRAY`.

```suggestion
    //! \brief Constructs one-dimensional range.
    NDRange(array<size_type, 1> a) : NDRange(a[0]){}

    //! \brief Constructs two-dimensional range.
    NDRange(array<size_type, 2> a) : NDRange(a[0], a[1]){}

    //! \brief Constructs three-dimensional range.
    NDRange(array<size_type, 3> a) : NDRange(a[0], a[1], a[2]){}
```

bashbaug (commented):
This PR currently adds non-member function overloads which work on the default command queue.  I think we should add member function overloads as well.

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Good point. Thank you!

Mikolaj-Filar-Mobica:
> The array changes look good, thanks!
> 
> I still think we need the member function overloads though.
Ahh, I hadn't noticed that comment. Done.

include/CL/opencl.hpp
```diff
@@ -4524,6 +4524,13 @@ class Image : public Memory
         }
         return param;
     }
+
+    Buffer getBuffer()
```
bashbaug, line 4528: I'm not following why we need `getBuffer` - can you say more?

Any shared functionality between buffers and images should go in the "memory" class, so we shouldn't ever need to go from an "Image" to a "Buffer".

include/CL/opencl.hpp
```diff
@@ -8199,6 +8370,24 @@ class CommandQueue : public detail::Wrapper<cl_command_queue>
         return result;
     }
 
+     inline void* enqueueMapImage(
+         const Image& buffer,
```
bashbaug, line 8374: Nitpick: Since this is an "Image" argument I think it's a little confusing to call it a "buffer".

```suggestion
         const Image& image,
```

(We'll need to change where it's used below, also.)

keryell:
Why adding such a specific support for `std::array`?
Why not adding just support for ranges, i.e. anything dealing with `std::begin()` and `std::end()?

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: I thought I needed a `cl::Buffer` to work with a `cl::KernelFunctor` in the blur example, but after your comment I see I can just use a `cl::Memory` with the same effect as well. Thanks for the comment. I've removed the `getBuffer` method.

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: Good point! Done.

Mikolaj-Filar-Mobica:
> Why adding such a specific support for `std::array`? Why not adding just support for ranges, i.e. anything dealing with `std::begin()` and `std::end()?

I've added it here because 2D ranges can be useful for processing 2D objects, which I think is common (image, matrix, etc.). Of course, anyone can just extend it manually as it's done in these wrappers. I don't see much benefit in creating more complex syntax with templates etc. for such a basic purpose. What do you think?

keryell:
> > Why adding such a specific support for `std::array`? Why not adding just support for ranges, i.e. anything dealing with `std::begin()` and `std::end()?
> 
> @keryell, I've added it here because 2D ranges can be useful for processing 2D objects, which I think is common (image, matrix, etc.). Of course, anyone can just extend it manually as it's done in these wrappers. I don't see much benefit in creating more complex syntax with templates etc. for such a basic purpose. What do you think?

I think this is C++ code and we should avoid adding more technical debt with more old C style extensive programming here instead of just writing generic reasonable™ C++.

What is the problem with templates or auto? We are not talking about code visible from the end-user but efficient library and generic code hidden in the library.

The idea here is to generalize your code taking `std::array` into taking any range, not adding yet-another-constructor for any new fancy object we can think of. Why not C arrays? Why not `std::list` then? I want `std::vector`, etc.

Instead of using `[1]` etc. you will need to use the more general `std::advance`.

Mikolaj-Filar-Mobica:
> > > Why adding such a specific support for `std::array`? Why not adding just support for ranges, i.e. anything dealing with `std::begin()` and `std::end()?
> > 
> > 
> > @keryell, I've added it here because 2D ranges can be useful for processing 2D objects, which I think is common (image, matrix, etc.). Of course, anyone can just extend it manually as it's done in these wrappers. I don't see much benefit in creating more complex syntax with templates etc. for such a basic purpose. What do you think?
> 
> I think this is C++ code and we should avoid adding more technical debt with more old C style extensive programming here instead of just writing generic reasonable™ C++.
> 
> What is the problem with templates or auto? We are not talking about code visible from the end-user but efficient library and generic code hidden in the library.
> 
> The idea here is to generalize your code taking `std::array` into taking any range, not adding yet-another-constructor for any new fancy object we can think of. Why not C arrays? Why not `std::list` then? I want `std::vector`, etc.
> 
> Instead of using `[1]` etc. you will need to use the more general `std::advance`.
I thought the code had to be C++11 compliant. I see that `std::advance` is a C++17 feature (thx, didn't know that). Also, it will contain some implementation instead of raw wrapping. Of course I can implement it this way, but it would be great to know the approach to know how to port/implement next things.

keryell:
> It will contain some implementation instead of raw wrapping. Of course I can implement it this way, but it would be great to know if the performance is not a crucial here.

Sorry, I do not understand this conversation.

Mikolaj-Filar-Mobica:
> Sorry, I do not understand this conversation. I am new again to this project. :-( Are you implementing some design goals from Khronos we could look at?
No, I don't implement design goals, but I can't understand yours. I have no idea how to implement it, using your suggestions to produce code that is easy to understand and without a few extra operations. I am open to a different solution. Could you please provide the code of your solution?

keryell (approved):
That looks good.
Generalizing with old C++11 would be too complex.
Providing constructor for `std::array` is still quite better than nothing.

include/CL/opencl.hpp
```diff
@@ -7762,6 +7771,35 @@ class CommandQueue : public detail::Wrapper<cl_command_queue>
         return err;
     }
 
+    inline cl_int enqueueReadBufferRect(
```
bashbaug, line 7774: Minor: an explicit `inline` is not required for any member functions defined in the class body.  For consistency and clarity I think we should remove it, both here and for all of the other added member functions in the class body.  (Note, the non-member functions added outside of the class body should still explicitly be marked `inline`!)

```suggestion
    cl_int enqueueReadBufferRect(
```

bashbaug (approved):
LGTM, just one minor comment that would be nice to fix since this PR needs a rebase now anyhow.

include/CL/opencl.hpp
Mikolaj-Filar-Mobica: That's absolutely true. I've removed unnecessary inlines and resolved conflicts. Thanks!

PR #203, opened 2023-03-13
add static assert for sizeof Event and cl_event

Lukasz-Krakowiak-Mobica:
static_assert for sizeof cl::Event and cl_event to be sure that both are equal

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -3395,6 +3395,9 @@ public:
     static cl_int
     waitForEvents(const vector<Event>& events)
     {
+        static_assert(sizeof(cl::Event) == sizeof(cl_event),
+        "Size of cl::Event must be equal to size of cl_event");
+
         return detail::errHandler(
             ::clWaitForEvents(
                 (cl_uint) events.size(), (events.size() > 0) ? (cl_event*)&events.front() : NULL),


include/CL/opencl.hpp
```diff
@@ -3395,6 +3395,9 @@ class Event : public detail::Wrapper<cl_event>
     static cl_int
     waitForEvents(const vector<Event>& events)
     {
+        static_assert(sizeof(cl::Event) == sizeof(cl_event),
+        "Size of cl::Event must be equal to size of cl_event");
```
bashbaug, line 3399: Formatting nitpick (it would be great to find some clang-format settings that work for this file eventually):

```suggestion
        static_assert(sizeof(cl::Event) == sizeof(cl_event),
            "Size of cl::Event must be equal to size of cl_event");
```

bashbaug (commented):
Thanks, this is definitely an improvement.

Even though the `static_assert` will apply to the entire file it's a little odd to put it in the `waitForEvents` function.  Couple of options:

1. Put it everywhere we use code like `(cl_event*)&events.front()`.  Probably overkill.
2. Put it outside of the `cl::Event` class with a nice comment describing why the sizes must match.
3. Do nothing, it's good enough as it is.

I'm fine with (3), at least for now.  I'll leave this PR open another day or so just in case someone else wants to advocate for a different option.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Hi Ben, thanks for comment.
This PR was come as a result of your last comment (https://github.com/KhronosGroup/OpenCL-CLHPP/pull/199#discussion_r1131198010).

I am agree that putting assert everywhere where events.front() is using, is overkill, so I decided to put this in one place.
Maybe you are right that static_assert in one of the methods of Event class is a little odd, but from other hand it also uses this kind of casting and it is closly place where unexpecting changes could happen. 

If you still will want to move this assert, I can put in this part of code: https://github.com/KhronosGroup/OpenCL-CLHPP/blob/main/include/CL/opencl.hpp#L3453
What do you think ?

bashbaug: I'm fine with this as-is.  We can always move it later...

bashbaug (approved):

Issue #150, opened 2021-10-12
Return type of cl::Device::getInfo<CL_DEVICE_PLATFORM>()


MathiasMagnus:
I believe the return type of `cl::Device::getInfo<CL_DEVICE_PLATFORM>()` is erronously the naked C type `cl_platform_id` instead of `cl::Platform`. Other queries do even more magic, and `cl::Context::getInfo<CL_CONTEXT_DEVICES>()` returns an instance of `cl::vector<cl:Device>` which is easy to iterate/filter, etc. Because the API object CTORS are explicit, this trivial code fails to compile:
```c++
cl::Platform platform = device.getInfo<CL_DEVICE_PLATFORM>();
```
```
error C2440: 'initializing': cannot convert from 'cl::detail::param_traits<cl::detail::cl_device_info,4145>::param_type' to 'cl::Platform'
note: Constructor for class 'cl::Platform' is declared 'explicit'
```
Instead we have to resort to explicit CTOR invocation as such:
```
cl::Platform platform{ device.getInfo<CL_DEVICE_PLATFORM>() };
```
But it also forbids us to use the return type immediately for things such as:
```
device.getInfo<CL_DEVICE_PLATFORM>().getInfo<CL_PLATFORM_VENDOR>()
```
We have to back and insert the CTOR invocation mid-air, making code less readable.

I believe this is a bug and we should fix.

PR #204, opened 2023-03-15
cl::Program deduction improvement - instead of cl_platform_id

pwisniewskimobica:
As a solution of issue https://github.com/KhronosGroup/OpenCL-CLHPP/issues/150
I propose changing the type in macros from cl_platform_id to cl::Platform

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -725,6 +725,7 @@ namespace cl {
 #endif
 
     class Memory;
+    class Platform;
     class Program;
     class Device;
     class Context;
@@ -1322,7 +1323,7 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     F(cl_device_info, CL_DEVICE_AVAILABLE, cl_bool) \
     F(cl_device_info, CL_DEVICE_COMPILER_AVAILABLE, cl_bool) \
     F(cl_device_info, CL_DEVICE_EXECUTION_CAPABILITIES, cl_device_exec_capabilities) \
-    F(cl_device_info, CL_DEVICE_PLATFORM, cl_platform_id) \
+    F(cl_device_info, CL_DEVICE_PLATFORM, cl::Platform) \
     F(cl_device_info, CL_DEVICE_NAME, string) \
     F(cl_device_info, CL_DEVICE_VENDOR, string) \
     F(cl_device_info, CL_DRIVER_VERSION, string) \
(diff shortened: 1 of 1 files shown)


pwisniewskimobica:
To fix the error I added help method to Device class
Thanks

include/CL/opencl.hpp
```diff
@@ -2432,7 +2437,7 @@ class Device : public detail::Wrapper<cl_device_id>
         static PFN_clCreateSubDevicesEXT pfn_clCreateSubDevicesEXT = nullptr;
 #if CL_HPP_TARGET_OPENCL_VERSION >= 120
         cl::Device device(object_);
-        cl_platform_id platform = device.getInfo<CL_DEVICE_PLATFORM>();
+        cl_platform_id platform = this->getPlatformIdFromDevice(device);
```
bashbaug, line 2440: I think this will work but the helper function seems like overkill, especially since we're just calling `operator()` on line 10468 (see below).  Why don't we just call `operator()` here also?
```suggestion
        cl_platform_id platform = device.getInfo<CL_DEVICE_PLATFORM>()();
```

pwisniewskimobica: @bashbaug Do you have any other proposal?
That one which You proposed unfortunately seems is not compiling.
My one is compiling and working but I don't know why one of the CI jobs ran into timeout 
image

I think that the best other solution would be to put every derived class in separate  header - but it is quite big change.
The easiest way in my opinion is to use my change.
What do You think?
Thanks!

pwisniewskimobica: My other proposal is to move definition of `cl_int Device::createSubDevices(const cl_device_partition_property_ext* properties,
                        vector<Device>* devices)`
after the Platform class definition
This will also solve the issue.

pwisniewskimobica: I just putted this new proposal about removing definition of `createSubDevices` after Platform class

include/CL/opencl.hpp
```diff
@@ -2883,6 +2841,70 @@ class Platform : public detail::Wrapper<cl_platform_id>
...
+    typedef CL_API_ENTRY cl_int(CL_API_CALL * PFN_clCreateSubDevicesEXT)(
+        cl_device_id /*in_device*/,
+        const cl_device_partition_property_ext* /* properties */,
+        cl_uint /*num_entries*/, cl_device_id* /*out_devices*/,
+        cl_uint* /*num_devices*/) CL_API_SUFFIX__VERSION_1_1;
+
+    static PFN_clCreateSubDevicesEXT pfn_clCreateSubDevicesEXT = nullptr;
+#if CL_HPP_TARGET_OPENCL_VERSION >= 120
+    cl::Device device(object_);
+    cl_platform_id platform = this->getPlatformIdFromDevice(
```
bashbaug, line 2858: Since we moved the function definition after the definition of `cl::Platform` do we still need `getPlatformIdFromDevice`?

pwisniewskimobica: That's right. Just removed.
Thanks for that!

pwisniewskimobica:
It is failing in CI but to be honest right now I am investigating and analyzing why because it fails on line 10820 which in fact my change does not include.

include/CL/opencl.hpp
```diff
@@ -2466,93 +2467,10 @@ class Device : public detail::Wrapper<cl_device_id>
     }
 #endif // #if CL_HPP_TARGET_OPENCL_VERSION >= 210
 
-#if CL_HPP_TARGET_OPENCL_VERSION >= 120
-    //! \brief Wrapper for clCreateSubDevices().
-    cl_int createSubDevices(
-        const cl_device_partition_property * properties,
-        vector<Device>* devices)
```
bashbaug, line 2473: It looks like this version of the function (the OpenCL 1.2 version that uses `cl_device_partition_property`, not `cl_device_partition_property_ext`) was erroneously removed.  We need to keep it.  Note, I think it can stay here, in the class definition, because it doesn't need to get the platform to query extension pointers.

bashbaug (changes requested):
Thanks!  This is getting close.  One more issue though...

include/CL/opencl.hpp
pwisniewskimobica: that's right.
Just fixed.
Thanks!

bashbaug:
I'm working on a fix for the CI failures. I don't think they're related to these changes.

pwisniewskimobica:
> I can't seem to figure out if there's semantic difference in the big chunk of formatting changes. Adding those few extra preprocessor statements caused the braces and the entire code to reduce indentation. Can the changeset be made smaller? If not, then it's fine the way this way too, it's just strange that the meaningful change is 5 important lines, then it become +104, -86 lines.

pwisniewskimobica:
> I can't seem to figure out if there's semantic difference in the big chunk of formatting changes. Adding those few extra preprocessor statements caused the braces and the entire code to reduce indentation. Can the changeset be made smaller? If not, then it's fine the way this way too, it's just strange that the meaningful change is 5 important lines, then it become +104, -86 lines.

@MathiasMagnus thanks for reply
This big change is just because I separated the method definition from its declaration.
Its small change but in code diffs in looks quite big.
I needed to do this because without separating definition from its declaration then its code does not see  Platform class.
Regards

MathiasMagnus (approved):
Ship it!

bashbaug (approved):
Thanks, latest version LGTM!

This is a breaking change, meaning previously-written code would need to be modified after this change.  I think this is the right thing to do, just calling it out for completeness.

The kind of code that will need to change is:

```c++
// this will no longer work:
cl_platform_id platform = my_device.getInfo<CL_DEVICE_PLATFORM>();
```

The fix is to add an extra set of parentheses to explicitly call `operator()`:

```c++
// fix is to call operator():
cl_platform_id platform = my_device.getInfo<CL_DEVICE_PLATFORM>()();
```

bashbaug:
Merging as discussed in the May 2nd teleconference.

PR #205, opened 2023-03-16
Ignore deprecated declarations in tests also for MSVC compiler

Mikolaj-Filar-Mobica:
This exists for GCC, so I've extended it to the Microsoft compiler so that the logs have fewer warnings.
Currently, the build log contains more than 9k warnings, and about 7k of them are about deprecation.


bashbaug (approved):
Thanks, this is a nice improvement!

Issue #206, opened 2023-03-16
Semaphore support break compilation with clang


serjzimmerman:
Currently the headers can't be used with clang (tested with 14.0.6 and 15.0.7) when ```cl_khr_semaphore``` is defined.
The culprit is d82dce3. The code indeed should not compile because of the following:

https://github.com/KhronosGroup/OpenCL-CLHPP/blob/8c5374baa80604cdd1a2d4c7a979b91bca7e0a9b/include/CL/opencl.hpp#L8946-L8954

Here ```Semaphore``` is an incomplete type and ```std::vector<Semaphore>::size()``` is called:

```cpp
err = pfn_clEnqueueWaitSemaphoresKHR(
                    ....
                    (cl_uint)sema_objects.size(),
                    ....);
```

This issue is similar to https://github.com/KhronosGroup/OpenCL-CLHPP/pull/175.

I found that the issue appears only when using ```std=c++20```

Minimal steps to reproduce:

CMakeLists.txt:
```cmake
add_executable(bug bug.cc)
target_link_libraries(bug PUBLIC OpenCL::OpenCL OpenCL::Headers
                                              OpenCL::HeadersCpp)
target_compile_features(bug PRIVATE cxx_std_20)
```

bug.cc
```cpp
#include "CL/opencl.hpp"
int main() {}
```

Seems like the obvious solution is to move the ```enqueueWaitSemaphores``` definiton after the definiton of ```Semaphore```. Will create a PR that fixes this bug

bashbaug:
> Will create a PR that fixes this bug

Thanks!

Seems like we should also look into enabling C++20 compilation in our CI...

PR #175, opened 2022-04-06
Fix incomplete type in BuildLogType

MathiasMagnus:
Fixes #167

The gist of the issue is that `cl::Device` is still an incomplete type during the definition of `cl::BuildError` which causes a failure in instantiating `std::pair<cl::Device, cl::string>` which is part of `cl::BuildError`. Moving the definition of this type _after_ the definition of `cl::Device` solves the issue.

_(The implementation of `std::pair` in C++20 may not tolerate having incomplete types (I'm surprised it worked all along). Mind you we're not testing C++20 builds yet, but we should address that in a different PR.)_

PR #207, opened 2023-03-16
Fix Semaphore incomplete type

serjzimmerman:
Fixes https://github.com/KhronosGroup/OpenCL-CLHPP/issues/206.

Basically, ```Semaphore``` was used in a context where its type was required to be complete.
This is easily solved by moving the ```enqueueWaitSemaphores``` function body after the ```Semaphore``` definition.

Seems like gcc/clang disagree whether this was legal. The C++20 Standard agrees with clang:

[vector.overview]
An incomplete type T may be used when instantiating vector if the allocator meets the allocator completeness requirements (16.5.3.5.1). T shall be complete before any member of the resulting specialization of vector is referenced.

At some point CI should be expanded to cover newer standards to avoid issues like this and https://github.com/KhronosGroup/OpenCL-CLHPP/pull/175

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -8948,57 +8948,13 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
         const vector<Semaphore> &sema_objects,
         const vector<cl_semaphore_payload_khr> &sema_payloads = {},
         const vector<Event>* events_wait_list = nullptr,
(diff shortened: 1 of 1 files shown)


keryell:
> At some point CI should be expanded to cover newer standards to avoid issues like this and #175

Yes, it would be nice to check with all the various standards up C++23 with various compilers, when available.

Mikolaj-Filar-Mobica:
Thanks for the fix! I agree with you and @keryell that expanding the CI would be beneficial.

bashbaug (approved):

PR #208, opened 2023-03-21
remove methods earlier required by msvc

pwisniewskimobica:
remove explicit declared copy constructors/copy assigments , 
move constructors etc which earlier seems to be required by msvc


Mikolaj-Filar-Mobica (approved):

bashbaug:
Merging as discussed in the March 28th teleconference.

PR #209, opened 2023-03-24
C++ wrapper for clGetImageRequirementsInfoEXT

### LINKED ISSUE #176: Add support for cl_ext_image_requirements_info and cl_ext_image_from_buffer

Lukasz-Krakowiak-Mobica:
Closes #176


include/CL/opencl.hpp
```diff
@@ -4534,6 +4543,24 @@ class BufferRenderGL : public Buffer
     }
 };
 
+#if defined(cl_ext_image_requirements_info)
+enum class ImgReqInfoExt : cl_image_requirements_info_ext
+{
+    ImgReqNone = 0,
+    ImgReqRowPitchAlign = CL_IMAGE_REQUIREMENTS_ROW_PITCH_ALIGNMENT_EXT,
```
bashbaug, line 4550: I think we should rethink this naming convention slightly since the current full enum name will be `ImgReqInfoExt::ImgReqNone`, which is a little redundant.

Could this just be:

`ImgReqInfo::NoneExt`?

* Don't need to duplicate `ImgReq` on the enum name itself when it is already in the enum class name.
* Move `Ext` to the enum name rather than the enum class name?
* We might also want to consider spelling out `Image` or `Requirements` (or both)?

include/CL/opencl.hpp
```diff
@@ -4618,6 +4645,41 @@ class Image : public Memory
         }
         return param;
     }
+
+#if defined(cl_ext_image_requirements_info)
+    static cl_int getImageRequirementsInfoExt(const Context &context,
```
bashbaug, line 4650: Debatably this should be a member function of the `Context` similar to `getSupportedImageFormats`.

If we're going to keep it as a static function though I don't think we need "Image" in the name since it would be redundant: `Image::getImageRequirementsInfoExt`

include/CL/opencl.hpp
```diff
@@ -4618,6 +4645,41 @@ class Image : public Memory
...
+        cl_mem_flags flags,
+        ImgReqInfoExt param_name,
+        size_type param_value_size,
+        void* param_value,
+        size_type* param_value_size_ret,
+        const cl_image_format* image_format = nullptr,
+        const cl_image_desc* image_desc = nullptr,
+        const cl_mem_properties* properties = nullptr)
```
bashbaug, line 4658: I'm not going to lose much sleep over this but it's a little odd to have these parameters with default values in this specific order, and I wonder if it makes more sense to have specific overloads instead.

bashbaug (commented):
Could you include a test for this functionality?

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: I felt that clGetSupportedImageFormats is rather more general information and clGetImageRequirementsInfoEXT api is more closer to Image clas,
But if you will have other idea, I can move it to Context.

include/CL/opencl.hpp
```diff
@@ -4530,6 +4539,24 @@ class BufferRenderGL : public Buffer
     }
 };
 
+#if defined(cl_ext_image_requirements_info)
+enum class ImageRequirementsInfoExt : cl_image_requirements_info_ext
+{
+    ExtNone = 0,
```
bashbaug, line 4545: The new version is much better - thanks! - but there's still a small amount of redundancy since the full name will be:

`ImageRequirementsInfoExt::ExtNone`

Options:

1. Do nothing, leave it as it is.
2. Take `Ext` off of the enum class name: `ImageRequirementsInfo::ExtNone`.
3. Take `Ext` off of the enum name: `ImageRequirementsInfoExt::None`.

I have a slight preference for (3).  What do you think?

include/CL/opencl.hpp
```diff
@@ -4586,6 +4613,73 @@ class Image : public Memory
         }
         return param;
     }
+
+#if defined(cl_ext_image_requirements_info)
+    static cl_int getRequirementsInfoExt(const Context &context,
```
bashbaug, line 4618: It still bugs me a little that this is a static function of the `Image` class whereas `getSupportedImageFormats` is a member function of the `Context` class, but it's mostly the inconsistency that bugs me.

Maybe we should consider adding a static function for getSupportedImageFormats and a member function for getRequirementsInfoExt?  There's no reason we can't have both functions in both places.

include/CL/opencl.hpp
```diff
@@ -4586,6 +4613,73 @@ class Image : public Memory
...
+                param_value_size_ret),
+                __GET_IMAGE_REQUIREMENT_INFO_EXT_ERR);
+    }
+
+static cl_int getRequirementsInfoExt(const Context &context,
+        cl_mem_flags flags,
+        ImageRequirementsInfoExt param_name,
+        size_type param_value_size,
+        void* param_value,
+        size_type* param_value_size_ret,
```
bashbaug, line 4656: Sorry I didn't notice this before: It'd be best to avoid having these kinds of param_value_size, param_value, and param_value_size_ret parameters in the C++ bindings.  Can we do something more like this instead?

https://github.com/KhronosGroup/OpenCL-CLHPP/pull/225#discussion_r1166271210

Basically, we'd need one variant that is templated on the info being queried, and another version that takes the info as an argument.  In both cases we shouldn't need void pointers, though.

bashbaug (commented):
This is coming along nicely - thanks!

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: I will move it to the Context class.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: @bashbaug  I am agree, best would be to use detail::getInfo method, but still there are a few more args:
image_format, image_desc, (properties) which should be forwarded also to the getInfoHelper in a proper order.

To solve this, I need to define own version of getInfoHelper function.
What do you think ?

bashbaug: Please have a look at the wrapper around `clGetKernelWorkGroupInfo` - it also needs to pass an additional device argument.  I think you could do something similar for this wrapper?

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Done thanks !

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Yes I saw this, but still there wasn't this what I would expected.
I create helper method: getImageRequirementsInfoExtHelper

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Done thanks !

tests/test_openclhpp.cpp
```diff
@@ -4115,4 +4119,105 @@ void testEnqueueAcquireExternalMemObjects(void) {}
...
+    auto info5 = context.getImageRequirementsInfoExt<cl::ImageRequirementsInfoExt::MaxHeight>();
+    TEST_ASSERT_EQUAL(info5, TEST_UINT_VALUE);
+
+    auto info6 = context.getImageRequirementsInfoExt<cl::ImageRequirementsInfoExt::MaxDepth>();
+    TEST_ASSERT_EQUAL(info6, TEST_UINT_VALUE);
+
+    auto info7 = context.getImageRequirementsInfoExt<cl::ImageRequirementsInfoExt::MaxArraySize>();
+    TEST_ASSERT_EQUAL(info7, TEST_UINT_VALUE);
+}
+#endif
```
bashbaug, line 4221: This needs an `#else` so the symbol is still defined even if the extension is not in the headers.

```suggestion
#else
void testTemplateGetImageRequirementsInfo() {}
#endif
```

bashbaug (approved):
LGTM, just have one suggestion to improve robustness with standard headers (we should have a  CI config for this).

tests/test_openclhpp.cpp
Lukasz-Krakowiak-Mobica: done :)

bashbaug:
Merging as discussed in the June 13th teleconference.

include/CL/opencl.hpp
```diff
@@ -3088,6 +3117,25 @@ UnloadCompiler()
...
 #endif // #if defined(CL_USE_DEPRECATED_OPENCL_1_1_APIS)
 
+
+#if defined(cl_ext_image_requirements_info)
+enum ImageRequirementsInfoExt : cl_image_requirements_info_ext
+{
+    RowPitchAlign = CL_IMAGE_REQUIREMENTS_ROW_PITCH_ALIGNMENT_EXT,
+    BaseAddAlign = CL_IMAGE_REQUIREMENTS_BASE_ADDRESS_ALIGNMENT_EXT,
+    Size = CL_IMAGE_REQUIREMENTS_SIZE_EXT,
+    NaxWidth = CL_IMAGE_REQUIREMENTS_MAX_WIDTH_EXT,
```
Anteru, line 3127: Typo: NaxWidth -> MaxWidth

PR #212, opened 2023-03-29
Tests for Device::GetInfo

pwisniewskimobica:
This PR is replacement for old PR [211](https://github.com/KhronosGroup/OpenCL-CLHPP/pull/211) which I removed
Tests for
Device::GetInfo<CL_DEVICE_NAME>
Device::GetInfo<CL_DEVICE_VENDOR_ID>
Device::GetInfo<CL_DEVICE_IMAGE_SUPPORT>

### DIFF
--- a/tests/test_openclhpp.cpp
+++ b/tests/test_openclhpp.cpp
@@ -71,9 +71,7 @@ static inline cl_semaphore_khr make_semaphore_khr(int index)
 }
 #endif
 
-/* Pools of pre-allocated wrapped objects for tests. There is no device pool,
- * because there is no way to know whether the test wants the device to be
- * reference countable or not.
+/* Pools of pre-allocated wrapped objects for tests.
  */
 static const int POOL_MAX = 5;
 static cl::Platform platformPool[POOL_MAX];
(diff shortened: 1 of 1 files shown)


tests/test_openclhpp.cpp
```diff
@@ -405,6 +406,7 @@ void setUp()
 #if defined(cl_khr_command_buffer)
         commandBufferKhrPool[i]() = make_command_buffer_khr(i);
 #endif
+        devicePool[i]() = make_device_id(i);
```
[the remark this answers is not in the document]
pwisniewskimobica: @Lukasz-Krakowiak-Mobica 
thanks for comments. Every of them applied

tests/test_openclhpp.cpp
```diff
@@ -75,6 +75,7 @@ static cl::Program programPool[POOL_MAX];
 #if defined(cl_khr_command_buffer)
 static cl::CommandBufferKhr commandBufferKhrPool[POOL_MAX];
 #endif
+static cl::Device devicePool[POOL_MAX];
```
bashbaug, line 78: This comment a few lines above makes me a little nervous:

```c
/* Pools of pre-allocated wrapped objects for tests. There is no device pool,
 * because there is no way to know whether the test wants the device to be
 * reference countable or not.
 */
```

Is this still relevant?  Do we really need a `devicePool`?

bashbaug (commented):
The tests themselves look good - thanks!

See my one comment about the `devicePool` addition.

tests/test_openclhpp.cpp
pwisniewskimobica: @bashbaug Thanks for reply.
Comment just updated- because we have now devicePool and we need it.

bashbaug (approved):
LGTM!

PR #213, opened 2023-03-29
Rename enqueueSignalSemaphore to enqueueSignalSemaphores for consistency

kamilwiatrowskimobica:
Rename `CommandQueue::enqueueSignalSemaphore` to `CommandQueue::enqueueSignalSemaphores` to keep consistency with the underlying C API function.

There is `CommandQueue::enqueueWaitSemaphores()` as wrapper for `clEnqueueWaitSemaphoresKHR` so I assume it should be `CommandQueue::enqueueSignalSemaphores()` for `clEnqueueSignalSemaphoresKHR`.

It seems to be a recent addition so I hope it is not considered as a break of backwards compatibility.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -8665,7 +8665,7 @@ typedef CL_API_ENTRY cl_int (CL_API_CALL *PFN_clEnqueueReleaseD3D10ObjectsKHR)(
         const vector<Event>* events_wait_list = nullptr,
         Event *event = nullptr) const;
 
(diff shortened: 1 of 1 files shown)


Mikolaj-Filar-Mobica (approved):

bashbaug (approved):
LGTM!  It makes sense to fix this sooner rather than later, so merging ASAP.

bashbaug:
I'm going to go ahead and merge this.  If there are any CI failures (there shouldn't be) we can address them after the fact.

PR #214, opened 2023-04-02
fix CI testing for MacOS and Linux

bashbaug:
This PR applies the same CI fixes from other repos to OpenCL-CLHPP:

- It removes the Ubuntu 18.04 configs because the Ubuntu 18.04 runner is deprecated.
- Explicitly installs gcc-9 on MacOS since it is no longer installed by default.
- Removes testing from OpenCL-Headers and OpenCL-ICD-Loader to speed up the builds.

I haven't moved any of the testing for Ubuntu 18.04 to Ubuntu 20.04 - this can be done in a separate PR.


MathiasMagnus (approved):

MathiasMagnus:
Not disabling testing of the deps was a grave error from my part.

bashbaug:
Closing in favor of #216.

PR #216, opened 2023-04-03
Ci fix

Kerilk:
This should fix the CI.

 - It installs missing compiler on MacOS,
 - It sets -DBUILD_TESTING=OFF for all platforms for the headers and loader,
 - It uses a runner to check for minimum environment to bypass Ubuntu 18.04 deprecation by github.

@MathiasMagnus I replaced absolute paths when possible by the relevant environment variables. The only exception is here:
https://github.com/Kerilk/OpenCL-CLHPP/blob/8fa5b4ef8ab067d17c9bd7577ede471e5b3ef552/.github/workflows/linux.yml#L25-L26

But I think the core of the issue is in https://github.com/actions/runner/issues/2058 and once GitHub fixes it we should be able to remove these two lines.

Please tell me if you had something else in mind.

This is a less drastic change than what is proposed here: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/214

### DIFF
--- a/.github/workflows/linux.yml
+++ b/.github/workflows/linux.yml
@@ -307,7 +307,7 @@ jobs:
     - name: "Consume (SDK): Configure/Build/Test"
       shell: bash
       run: $CMAKE_EXE -E make_directory $GITHUB_WORKSPACE/install/share/cmake/OpenCL ;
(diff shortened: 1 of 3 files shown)


bashbaug (approved):
LGTM - this is better than what I did!

There is still one place where we enable testing with `-D BUILD_TESTS=ON`.  Is this a typo, and should this be `-D BUILD_TESTING=ON` instead?

I think testing is enabled by default so this _probably_ isn't causing a problem in practice.

MathiasMagnus (approved):

Kerilk:
Thanks for the reviews!

> There is still one place where we enable testing with `-D BUILD_TESTS=ON`. Is this a typo, and should this be `-D BUILD_TESTING=ON` instead?

I think it is and implemented the change (after some fumble on my part). I'll check that the runs are indeed correct.

Kerilk:
I checked the CI run and it tested well with the latest modifications. This one is ready for merging I think.

Issue #217, opened 2023-04-03
`cl_image_desc desc = {0}` is not valid C++ zero initialization.


Kerilk:
I noticed quite a lot of warnings in the logs for the C++ bindings related to `cl_image_desc` initialization (or lack rather):

```
In file included from /Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/examples/src/trivialSizeTCompat.cpp:5:
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp: In constructor 'cl::Image1D::Image1D(const cl::Context&, cl_mem_flags, cl::ImageFormat, cl::size_type, void*, cl_int*)':
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::image_width' [-Wmissing-field-initializers]
 4539 |         cl_image_desc desc = {0};
      |                                ^
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::image_height' [-Wmissing-field-initializers]
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::image_depth' [-Wmissing-field-initializers]
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::image_array_size' [-Wmissing-field-initializers]
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::image_row_pitch' [-Wmissing-field-initializers]
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::image_slice_pitch' [-Wmissing-field-initializers]
/Users/runner/work/OpenCL-CLHPP/OpenCL-CLHPP/include/CL/opencl.hpp:4539:32: warning: missing initializer for member '_cl_image_desc::num_mip_levels' [-Wmissing-field-initializers]
[...]

PR #218, opened 2023-04-04
Fix zero initialization of cl_image_desc.

Kerilk:
This fixes https://github.com/KhronosGroup/OpenCL-CLHPP/issues/217 by using the correct syntax for the structure initialization, default values for integer and pointers being 0 and `nullptr` this should be equivalent to the intended zero initialization.


bashbaug (approved):
LGTM!

I'll wait another day or so before merging just in case @keryell has comments.

PR #219, opened 2023-04-05
add UTs for Kernel::setArg uses clSetKernelArgSVMPointer

Lukasz-Krakowiak-Mobica:


tests/test_openclhpp.cpp
```diff
@@ -1665,6 +1665,33 @@ void testKernelSetArgLocal()
...
+    VECTOR_CLASS<int> vec(1000);
+    clSetKernelArgSVMPointer_ExpectAndReturn(make_kernel(1), 2, vec.data(), CL_SUCCESS);
+    TEST_ASSERT_EQUAL(kernelPool[1].setArg(2, vec), CL_SUCCESS);
+#endif
+}
+
+void testKernelSetArgBySetKernelArgSVMPointerWithPointerType()
+{
+#if CL_HPP_TARGET_OPENCL_VERSION >= 200
+    cl_mem *memory = &bufferPool[1]();
```
bashbaug, line 1689: This could also pass any random pointer here too, including e.g. the address of a stack variable, but this should be fine too.

tests/test_openclhpp.cpp
```diff
@@ -1665,6 +1665,33 @@ void testKernelSetArgLocal()
     kernelPool[0].setArg(2, cl::Local(123));
 }
 
+void testKernelSetArgBySetKernelArgSVMPointerWithUniquePtrType()
+{
+#if CL_HPP_TARGET_OPENCL_VERSION >= 200
+    std::unique_ptr<int> buffer(new int(1000));
```
bashbaug, line 1671: For the record, debatably this should be `cl::Pointer` instead, especially since we're using `VECTOR_CLASS` for the vector type case, but this needs to pass an explicit `Deleter` so I don't think it's worth it.

bashbaug (approved):

PR #220, opened 2023-04-05
Unit tests for cl::Semaphore

kamilwiatrowskimobica:
Fix pointer param for clGetSemaphoreInfoKHR.
Add more params supported by clGetSemaphoreInfoKHR.

### DIFF
--- a/tests/test_openclhpp.cpp
+++ b/tests/test_openclhpp.cpp
@@ -3431,6 +3431,7 @@ void testMoveConstructSemaphoreNonNull(void) {}
 void testMoveConstructSemaphoreNull(void) {}
 #endif
 
+#if defined(cl_khr_semaphore)
 static cl_int clEnqueueWaitSemaphoresKHR_testEnqueueWaitSemaphores(
     cl_command_queue command_queue,
     cl_uint num_sema_objects,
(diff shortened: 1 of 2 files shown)


tests/test_openclhpp.cpp
```diff
@@ -3378,4 +3415,395 @@ void testCommandBufferInfoKHRCommandQueues()
...
+void testMoveAssignSemaphoreNull(void) {}
+void testMoveConstructSemaphoreNonNull(void) {}
+void testMoveConstructSemaphoreNull(void) {}
+#endif
+
+static cl_int clEnqueueWaitSemaphoresKHR_testEnqueueWaitSemaphores(
+    cl_command_queue command_queue,
+    cl_uint num_sema_objects,
+    const cl_semaphore_khr* sema_objects,
+    const cl_semaphore_payload_khr* sema_payload_list,
```
bashbaug, line 3438: It'd be best to put all of these new tests inside of one big `#if defined(cl_khr_semaphore)`.

If the headers don't support semaphore (and hence `cl_khr_semaphore` is not defined) then `cl_semaphore_khr` and other types that these tests use won't be defined, either.

kamilwiatrowskimobica: Done.

bashbaug (approved):

PR #221, opened 2023-04-07
Fix warnings

Kerilk:
This fixes several classes of warnings that are creating a lot of noise in the build. Those may be hiding issues we would like to fix. I will rebase on https://github.com/KhronosGroup/OpenCL-CLHPP/pull/218 as I didn't want to muddy the waters by adding those commit to the existing PR. The classes of issues fixed are:
 - Explicitly cast unused arguments to `(void)`;
 - Cast `cl_uint` used as loop bounds to `(int)` to compare against the loop counter;
 - Initialize an array to 0;
 - Remove unused variable;
 - Remove extra `;` when unneeded after macros;
 - Add guards around static functions so that they are defined only when used.

Once this and https://github.com/KhronosGroup/OpenCL-CLHPP/pull/218 are merged, remaining warnings need to be investigated, and we may think about activating `-Werror` at least on Linux and MacOS targets.


bashbaug (approved):
Thanks, LGTM!  Still a few more but this is a nice improvement...

PR #222, opened 2023-04-09
add guards for cl_khr_command_buffer_mutable_dispatch

bashbaug:
Guard support for `cl_khr_command_buffer_mutable_dispatch` with the mutable dispatch defines rather than by `cl_khr_command_buffer`.  Some headers (like the latest Ubuntu 22.04 header package) have command buffer support but not mutable dispatch support.


EwanC (approved):

bashbaug:
Merging as discussed in the April 11th teleconference.

PR #224, opened 2023-04-11
reduce warnings in test_openclhpp.cpp file

pwisniewskimobica:
I done some fixes in test_openclhpp.cpp file 
due to reduce warnings

### DIFF
--- a/tests/test_openclhpp.cpp
+++ b/tests/test_openclhpp.cpp
@@ -733,7 +733,7 @@ static cl_context clCreateContext_testContextNonNullProperties(
 
     TEST_ASSERT_NOT_NULL(properties);
     TEST_ASSERT_GREATER_THAN(0, num_devices);
-    for (int i = 0; i < (int)num_devices; i++) {
+    for (unsigned int i = 0; i < num_devices; i++) {
         TEST_ASSERT_EQUAL(make_device_id(i), devices[i]);
     }
     if (errcode_ret != nullptr)
@@ -3341,7 +3341,7 @@ static cl_program clLinkProgram_testLinkProgram(cl_context context,
     TEST_ASSERT_EQUAL(device_list, nullptr);
     TEST_ASSERT_EQUAL(options, nullptr);
     TEST_ASSERT_NOT_EQUAL(num_input_programs, 0);
-    for (int i=0; i<(int)num_input_programs; i++)
+    for (unsigned int i=0; i<num_input_programs; i++)
         TEST_ASSERT_EQUAL_PTR(input_programs[i], make_program(i));
     TEST_ASSERT_EQUAL(pfn_notify, nullptr);
     TEST_ASSERT_EQUAL(user_data, nullptr);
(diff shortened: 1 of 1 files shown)


tests/test_openclhpp.cpp
```diff
@@ -1561,7 +1561,7 @@ static cl_mem clCreateImage3D_testCreateImage3D_1_1(
     TEST_ASSERT_EQUAL(16, image_depth);
     TEST_ASSERT_EQUAL(256, image_row_pitch);
     TEST_ASSERT_EQUAL(65536, image_slice_pitch);
-    TEST_ASSERT_EQUAL_PTR((void *) 0xdeadbeef, host_ptr);
+    TEST_ASSERT_EQUAL_PTR(nullptr, host_ptr);
```
Mikolaj-Filar-Mobica, line 1564: I don't think changing `0xdeadbeef` to `nullptr` is a good approach. I see the better solution is suggested in https://github.com/KhronosGroup/OpenCL-CLHPP/pull/223 by @Kerilk:
`    TEST_ASSERT_EQUAL_PTR((void *)(size_t)0xdeadbeef, host_ptr);
`
For this reason, this PR is also in conflict with the one mentioned above.

pwisniewskimobica: Thanks 
You are right - changed back from nullptr to (void *)(size_t)0xdeadbeef
Other thing is that https://github.com/KhronosGroup/OpenCL-CLHPP/pull/223 
is not the same changes. This one and that are not conflicting- They are another fixes.

bashbaug:
Is this still needed now that we merged https://github.com/KhronosGroup/OpenCL-CLHPP/pull/223?

pwisniewskimobica:
@bashbaug 
I can say that https://github.com/KhronosGroup/OpenCL-CLHPP/pull/223 is another set of fixes.
For my it looks like it is designed mostly for linux while my fixes are for windows

bashbaug:
I'm pretty sure Windows builds were also warning-free when we merged #223.  Here are the CI runs:

https://github.com/KhronosGroup/OpenCL-CLHPP/actions/runs/4799850369/jobs/8540025875

If we have additional warnings on Windows then they're either configurations we aren't testing (possible) or we've introduced new warnings recently (also possible).

pwisniewskimobica:
@bashbaug basically you and me are right;)
I mean that you are right because CI does not shows warnings that I have on my local and I am right because Its strange but- the only difference I see is that CI uses cmake version 3.22.0 and I use 3.26
To proove I am attaching print screen
image

Kerilk:
I think the confusion comes from this discrepancy between what is in the repository:
https://github.com/KhronosGroup/OpenCL-CLHPP/blob/0f420fb265d5d3a64d68deceabfd3689f7322e30/tests/test_openclhpp.cpp#L736
and what you have on your local machine (per your screenshot):
```
    for (int i = 0; i < num_devices; i++) {
```
So your rebase must only have been partial.

pwisniewskimobica:
@bashbaug Thats right
That case I think this PR is not needed anymore
thanks

PR #225, opened 2023-04-12
C++ wrapper for cl_khr_external_semaphore

### LINKED ISSUE #155: Add support for cl_khr_external_semaphore* extensions

Lukasz-Krakowiak-Mobica:
Closes #155

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -1543,6 +1543,19 @@ inline cl_int getInfoHelper(Func f, cl_uint name, T* param, int, typename T::cl_
     F(cl_device_info, CL_DEVICE_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR,      cl::vector<cl_external_semaphore_handle_type_khr>) \
     F(cl_semaphore_info_khr, CL_SEMAPHORE_EXPORT_HANDLE_TYPES_KHR,      cl::vector<cl_external_semaphore_handle_type_khr>) \
 
+#define CL_HPP_PARAM_NAME_CL_KHR_EXTERNAL_SEMAPHORE_DX_FENCE_EXT(F) \
+    F(cl_external_semaphore_handle_type_khr, CL_SEMAPHORE_HANDLE_D3D12_FENCE_KHR, void*) \
+
+#define CL_HPP_PARAM_NAME_CL_KHR_EXTERNAL_SEMAPHORE_OPAQUE_FD_EXT(F) \
+    F(cl_external_semaphore_handle_type_khr, CL_SEMAPHORE_HANDLE_OPAQUE_FD_KHR, int) \
+
+#define CL_HPP_PARAM_NAME_CL_KHR_EXTERNAL_SEMAPHORE_SYNC_FD_EXT(F) \
+    F(cl_external_semaphore_handle_type_khr, CL_SEMAPHORE_HANDLE_SYNC_FD_KHR, int) \
+
+#define CL_HPP_PARAM_NAME_CL_KHR_EXTERNAL_SEMAPHORE_WIN32_EXT(F) \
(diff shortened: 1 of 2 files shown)


include/CL/opencl.hpp
```diff
@@ -10463,6 +10505,28 @@ class Semaphore : public detail::Wrapper<cl_semaphore_khr>
         return param;      
     }
 
+#ifdef cl_khr_external_semaphore
+    cl_int getSemaphoreHandleForTypeKHR (Device device,
+        ExtSemaphoreHandleTypeKhr handle_type,
+        size_type handle_size,
+        void* handle_ptr,
+        size_type* handle_size_ret)
```
bashbaug, line 10513: This is OK, but I think I'd really want to write code more like this:

```c++
auto my_handle = my_semaphore->getHandleForTypeKHR<TypeD3D12Fence>(Device);
```

Note, this is similar to other queries, like the different `getInfo`, `getWorkGroupInfo`, and `getSubGroupInfo` queries.

What do you think?

include/CL/opencl.hpp
```diff
@@ -10400,6 +10421,27 @@ namespace compatibility {
...
 
 #ifdef cl_khr_semaphore
+
+#ifdef cl_khr_external_semaphore
+enum class ExtSemaphoreHandleTypeKhr : cl_external_semaphore_handle_type_khr
+{
+    TypeNone = 0,
+#ifdef cl_khr_external_semaphore_dx_fence
+    TypeD3D12Fence = CL_SEMAPHORE_HANDLE_D3D12_FENCE_KHR,
+#endif
```
bashbaug, line 10431: There is a little bit of naming redundancy here and some inconsistency with the recently added `ExternalMemoryType` enum.

How do you feel about:

```suggestion
enum class ExternalSemaphoreType : cl_external_semaphore_handle_type_khr
{
    None = 0,
#ifdef cl_khr_external_semaphore_dx_fence
    D3D12Fence = CL_SEMAPHORE_HANDLE_D3D12_FENCE_KHR,
#endif
```

etc.

Key changes are:
* Spell out `External`.
* Drop `Handle` and `Khr` from the enum type name.
* Drop `Type` from the enums themselves.

This means we will write `ExternalSemaphoreType::D3D12Fence` rather than `ExtSemaphoreHandleTypeKhr::TypeD3D12Fence`, which seems shorter and more descriptive.

include/CL/opencl.hpp
bashbaug: For completeness, I think we could also have a variant that takes a parameter for the handle type, for cases where the handle type is only known at runtime, but I still think we'd want it to be more like:

```c++
cl_int getHandleForType<T>(const Device& device, ExtSemaphoreHandleTypeKhr handle_type, T* handle_ptr);
```

This would let you write code like this that doesn't need to worry about `void*`s.

```c++
FILE fp;
my_semaphore->getHandleForTypeKHR(my_device, TypeOpaqueFd, &fp);
```

Note, there is a variant like this for `getInfo`, also.

Lukasz-Krakowiak-Mobica: I think it has a sense :)
Done, thanks.

include/CL/opencl.hpp
Lukasz-Krakowiak-Mobica: Done, thanks.

include/CL/opencl.hpp
bashbaug: Thanks, but this isn't quite what I was looking for.  Have a look at: https://github.com/Oskar-Plaska-Mobica/OpenCL-CLHPP/pull/1

bashbaug (approved):
LGTM, and I fixed the merge conflict.  Just need CI to finish...

PR #226, opened 2023-04-17
test for svm allocator

pwisniewskimobica:
I wrote some code to test allocate and deallocate API (clSVMAlloc, clSVMFree, clEnqueueSVMMap)

### DIFF
--- a/tests/test_openclhpp.cpp
+++ b/tests/test_openclhpp.cpp
@@ -4188,4 +4188,47 @@ void testEnqueueAcquireExternalMemObjects(void) {}
 void testEnqueueReleaseExternalMemObjects(void) {}
 #endif // cl_khr_external_memory
 
+/****************************************************************************
+ * Tests for allocate and deallocate API (clSVMAlloc, clSVMFree, clEnqueueSVMMap)
+ ****************************************************************************/
+
+int *testMemory;
+void *clSVMAllocARM_stubForMemoryAllocation(cl_context context,cl_svm_mem_flags flags,
+                         size_t size, cl_uint alignment, int cmock_num_calls)
+{
+    TEST_ASSERT_EQUAL_HEX(size, 3 * sizeof(int));
+    testMemory = new int[3];
+    return testMemory;
+}
+cl_int clEnqueueSVMMap_stubForMemoryAllocation(cl_command_queue command_queue, cl_bool blocking_map, cl_map_flags flags, void* svm_ptr, size_t size, cl_uint num_events_in_wait_list, const cl_event* event_wait_list, cl_event* event, int cmock_num_calls)
+{
+    TEST_ASSERT_EQUAL_PTR(svm_ptr, testMemory);
+    TEST_ASSERT_EQUAL_HEX(size, 3 * sizeof(int));
+    return CL_SUCCESS;
+}
+void clSVMFree_stubForMemoryAllocation(cl_context context, void *svm_pointer,
+    int cmock_num_calls)
+{
+    TEST_ASSERT_EQUAL_PTR(svm_pointer, testMemory);
+    delete svm_pointer; 
+}
+void testSVMMemoryAllocation()
+{
+
+#if CL_HPP_TARGET_OPENCL_VERSION >= 200
+    cl::SVMAllocator<int, cl::SVMTraitCoarse<>> svmAllocator;
+        
+    clSVMAlloc_StubWithCallback(clSVMAllocARM_stubForMemoryAllocation);
+    clEnqueueSVMMap_StubWithCallback(clEnqueueSVMMap_stubForMemoryAllocation);
+    clSVMFree_StubWithCallback(clSVMFree_stubForMemoryAllocation);
+
+    int *ptr = svmAllocator.allocate(3);
+    TEST_ASSERT_EQUAL_PTR(ptr,testMemory);
+    clSVMFree_Expect(0, ptr);
+    svmAllocator.deallocate(ptr,3);
+#endif
+}
+
+
+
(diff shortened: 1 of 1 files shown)


tests/test_openclhpp.cpp
```diff
@@ -3378,4 +3378,47 @@ void testCommandBufferInfoKHRCommandQueues()
...
+    TEST_ASSERT_EQUAL_PTR(svm_pointer, testMemory);
+    delete svm_pointer; 
+}
+void testSVMMemoryAllocation()
+{
+
+#if CL_HPP_TARGET_OPENCL_VERSION >= 200
+    cl::SVMAllocator<int, cl::SVMTraitCoarse<>> svmAllocator;
+        
+    clSVMAlloc_StubWithCallback(clSVMAllocARM_stubForMemoryAllocation);
```
bashbaug, line 3411: This is a little confusing - we shouldn't have `ARM` in the name of this function since we aren't stubbing the ARM version of this function.

```suggestion
    clSVMAlloc_StubWithCallback(clSVMAlloc_stubForMemoryAllocation);
```

tests/test_openclhpp.cpp
```diff
@@ -3378,4 +3378,47 @@ void testCommandBufferInfoKHRCommandQueues()
...
+{
+    TEST_ASSERT_EQUAL_PTR(svm_ptr, testMemory);
+    TEST_ASSERT_EQUAL_HEX(size, 3 * sizeof(int));
+    return CL_SUCCESS;
+}
+void clSVMFree_stubForMemoryAllocation(cl_context context, void *svm_pointer,
+    int cmock_num_calls)
+{
+    TEST_ASSERT_EQUAL_PTR(svm_pointer, testMemory);
+    delete svm_pointer;
```
bashbaug, line 3403: This generates a warning on my system:

```
/home/bashbaug/git/OpenCL-CLHPP/tests/test_openclhpp.cpp: In function ‘void clSVMFree_stubForMemoryAllocation(cl_context, void*, int)’:
/home/bashbaug/git/OpenCL-CLHPP/tests/test_openclhpp.cpp:3403:12: warning: deleting ‘void*’ is undefined [-Wdelete-incomplete]
 3403 |     delete svm_pointer;
```

tests/test_openclhpp.cpp
```diff
@@ -3378,4 +3378,47 @@ void testCommandBufferInfoKHRCommandQueues()
...
 
+/****************************************************************************
+ * Tests for allocate and deallocate API (clSVMAlloc, clSVMFree, clEnqueueSVMMap)
+ ****************************************************************************/
+
+int *testMemory;
+void *clSVMAllocARM_stubForMemoryAllocation(cl_context context,cl_svm_mem_flags flags,
+                         size_t size, cl_uint alignment, int cmock_num_calls)
+{
+    TEST_ASSERT_EQUAL_HEX(size, 3 * sizeof(int));
```
bashbaug, line 3389: Can we check some of the other parameters to this function also?

For example, the test explicitly requested `cl::SVMTraitCoarse`, so we should be able to check the flags.  Can we test the alignment too?

tests/test_openclhpp.cpp
```diff
@@ -3378,4 +3378,47 @@ void testCommandBufferInfoKHRCommandQueues()
 #endif
 }
 
+/****************************************************************************
+ * Tests for allocate and deallocate API (clSVMAlloc, clSVMFree, clEnqueueSVMMap)
+ ****************************************************************************/
+
+int *testMemory;
```
bashbaug, line 3385: Is there any way to do this test without needing a global variable?  I'm not going to lose a ton of sleep over it, but if we ever want to enable parallel test execution it will be important not to have global variables, and we've been pretty good about avoiding global variables so far.

Does this test even need to allocate any real memory?  If not, can we just return one of the e.g. platform handles (`make_platform_id(0)`)?

tests/test_openclhpp.cpp
```diff
@@ -3833,4 +3833,46 @@ void testSemaphoreRetain(void) {}
...
+#if CL_HPP_TARGET_OPENCL_VERSION >= 200
+    cl::SVMAllocator<int, cl::SVMTraitCoarse<>> svmAllocator;
+        
+    clSVMAlloc_StubWithCallback(clSVMAlloc_stubForMemoryAllocation);
+    clEnqueueSVMMap_StubWithCallback(clEnqueueSVMMap_stubForMemoryAllocation);
+    clSVMFree_StubWithCallback(clSVMFree_stubForMemoryAllocation);
+
+    int *ptr = svmAllocator.allocate(3);
+    TEST_ASSERT_EQUAL_PTR(ptr,testMemory);
+    clSVMFree_Expect(0, ptr);
```
bashbaug, line 3873: I can't figure out what this line is doing - is it needed?

bashbaug (commented):
Thanks!  Just one last comment.  Also needs a rebase (hopefully trivial).

tests/test_openclhpp.cpp
bashbaug: Sorry about that, this is a case where the GitHub UI didn't do me any favors.

I'm asking about this line specifically: `clSVMFree_Expect(0, ptr);`

We already have `clSVMFree_StubWithCallback` so is the additional `clSVMFree_Expect` needed?  The test seems to work fine when I remove it, and I can't figure out why `clSVMFree` would be called twice.

bashbaug (approved):
LGTM, just waiting for CI to complete (which has been a problem recently...).

bashbaug:
Merging as discussed in the May 30th teleconference.

PR #228, opened 2023-04-17
wrapper for clEnqueueSVMMemFill and clEnqueueSVMMemcpy

bananAshkar:
1. SVMAllocator::allocate automatically maps the newly allocated SVM buffer. This is redundant for output buffers.I added a new argument to the method – bool map = true – to control whether the method should map the buffer.

2.opencl.hpp doesn't include wrapper methods for clEnqueueSVMMemFill and clEnqueueSVMMemcpy , so I added them.


include/CL/opencl.hpp
```diff
@@ -3888,7 +3892,8 @@ class SVMAllocator {
      */
     pointer allocate(
         size_type size,
-        typename cl::SVMAllocator<void, SVMTrait>::const_pointer = 0)
+        typename cl::SVMAllocator<void, SVMTrait>::const_pointer = 0,
+        cl_bool map=true)
```
bashbaug, line 3896: Nitpick: this should probably be:

```suggestion
        bool map = true)
```

Alternatively it could be `cl_bool map = CL_TRUE`, but I expect most users will want to pass `false` rather than `CL_FALSE` if they want to suppress the mapping.

bashbaug (approved):
LGTM!

include/CL/opencl.hpp
bananAshkar: Done!

bashbaug:
Merging as discussed in the May 30th teleconference.

PR #229, opened 2023-04-18
Eliminate `throw()`; use C++11 `noexcept`

Quuxplusone:
This header already uses `nullptr` and `Args&&...`, plus the `#error` directive on line 519 that forces C++11. So it should be safe to use the `noexcept` keyword too.

Also, Rule-of-Zero the Error class, to avoid triggering Clang's `-Wdeprecated-copy-with-dtor`.


bashbaug (approved):
Thanks, this a nice improvement!

PR #230, opened 2023-04-18
Use v.begin() for vectors instead of ADL begin(v)

### LINKED ISSUE #197: A curious name collision bug
```
#include <CL/opencl.hpp>
int main(){}
```
This code will compile. However, this:
```
#include <CL/opencl.hpp>
int end = 0; //Which is a variable I have purpose to use. Name this varaible as begin cause same error.
int main(){}
```
Cannot compile. This is the most curious bug I have ever encountered. It gives VS C2064 "term does not evaluate to a function taking 1 arguments", and "prev": cannot find indentifier.

As I know, compiler process source file line by line and follows look-up rule. Anything happen below SHOULD not interfere things before. I am very curios that declaring a variable of my own can make a header file producing error. I would be very happy if anyone can explain how such thing could happen. As for the error it does not matter anymore. I have many ways to settle it.

More spefically, this line is 1110, something like this:
```
param->assign(begin(value), prev(end(value)));
```
Where it use something named `end()`, but this is nothing special, just using some `std::string` utilities `begin()` `end()`. Why this specific header error such thing? header `cl.hpp` also use such but it does not generate error with my own variable in global namespace `begin` or `end`.

Quuxplusone:
No functional change (since `begin(x)` on a `std::vector` will always find `std::begin` which dispatches back to `x.begin()` anyway). But member name lookup is a lot cheaper, more idiomatic, and also avoids the arcane problem reported in #197.

Also use `v.end() - 1` instead of `prev(end(v))`.

Fixes #197.

### DIFF
--- a/examples/src/headerexample.cpp
+++ b/examples/src/headerexample.cpp
@@ -115,7 +115,7 @@ int main(void)
     // Traditional cl_mem allocations
 
     std::vector<int> output(numElements, 0xdeadbeef);
-    cl::Buffer outputBuffer(begin(output), end(output), false);
+    cl::Buffer outputBuffer(output.begin(), output.end(), false);
     cl::Pipe aPipe(sizeof(cl_int), numElements / 2);
 
     // Default command queue, also passed in as a parameter
@@ -152,7 +152,7 @@ int main(void)
         error
         );
 
-    cl::copy(outputBuffer, begin(output), end(output));
+    cl::copy(outputBuffer, output.begin(), output.end());
 
     cl::Device d = cl::Device::getDefault();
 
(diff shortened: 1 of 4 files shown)


Mikolaj-Filar-Mobica (approved):

bashbaug (approved):
LGTM!

bashbaug:
I'm going to go ahead and merge this one - this is a nice improvement.

@keryell I'm interested in your comments though, especially since you were active on #197.  Is there anything more you think we should do here?

Thanks again for your contribution!

keryell:
Thanks, that looks good.
@Quuxplusone: are you using OpenCL nowadays? :-)

keryell:
Interesting serendipity. :-)
Thanks.

PR #234, opened 2023-05-25
UT for getObjectInfo

piotr-wozniak-mobica:


piotr-wozniak-mobica:
> Hi, this isn't building for me, on Windows or Linux:
> 
> ```
> 10>------ Build started: Project: test_openclhpp_300_CL_HPP_USE_IL_KHR, Configuration: Debug x64 ------
> 10>test_openclhpp.cpp
> 10>D:\git\OpenCL-CLHPP\tests\test_openclhpp.cpp(4262,26): error C2601: 'testgetObjectInfo': local function definitions are illegal
> 10>D:\git\OpenCL-CLHPP\tests\test_openclhpp.cpp(4257): message : this line contains a '{' which has not yet been matched
> 10>D:\git\OpenCL-CLHPP\tests\test_openclhpp.cpp(11): fatal error C1075: '{': no matching token found
> 10>Done building project "test_openclhpp_300_CL_HPP_USE_IL_KHR.vcxproj" -- FAILED.
> ```
> 
> Can you please also say a little more what this case is intending to test? Thanks!

I just created a test for function that was not tested but after mocking functions I realise that there is only one scenario to cover in UT and its just passing value from another mocked function. Additionally this PR introduce mocks for functions from cl_gl.h file that can be useful for other UTs.

tests/test_openclhpp.cpp
```diff
@@ -4336,4 +4337,52 @@ void testTemplateGetImageRequirementsInfo()
...
+static cl_mem clCreateFromGLBuffer_testgetObjectInfo(cl_context context,
+                                                     cl_mem_flags flags,
+                                                     cl_GLuint bufobj,
+                                                     cl_int *errcode_ret,
+                                                     int num_calls)
+{
+    (void)num_calls;
+
+    TEST_ASSERT_EQUAL_PTR(make_context(0), context);
+    TEST_ASSERT_EQUAL(0, flags);
```
bashbaug, line 4349: Could we also test that the passed-in `bufobj` matches what was passed to the `cl::BufferGL` constructor?

tests/test_openclhpp.cpp
```diff
@@ -4336,4 +4337,52 @@ void testTemplateGetImageRequirementsInfo()
...
+    clGetGLObjectInfo_StubWithCallback(clGetGLObjectInfo_testgetObjectInfo);
+    clCreateFromGLBuffer_StubWithCallback(
+        clCreateFromGLBuffer_testgetObjectInfo);
+    clReleaseMemObject_StubWithCallback(clReleaseMemObject_testgetObjectInfo);
+    cl::BufferGL buffer(contextPool[0], flags, bufobj, &err);
+
+    buffer.getObjectInfo(&type, &bufobj);
+
+    TEST_ASSERT_EQUAL_PTR(make_mem(0), buffer());
+    TEST_ASSERT_EQUAL(CL_SUCCESS, err);
```
bashbaug, line 4386: Could we do these checks immediately after creating `buffer`?  They are unrelated to the call to `buffer.getObjectInfo()`.

tests/test_openclhpp.cpp
```diff
@@ -4336,4 +4337,52 @@ void testTemplateGetImageRequirementsInfo()
...
+    cl_GLuint bufobj = 0;
+    cl_mem memobj = make_mem(0);
+    cl_gl_object_type type = CL_GL_OBJECT_TEXTURE2D_ARRAY;
+    clGetGLObjectInfo_StubWithCallback(clGetGLObjectInfo_testgetObjectInfo);
+    clCreateFromGLBuffer_StubWithCallback(
+        clCreateFromGLBuffer_testgetObjectInfo);
+    clReleaseMemObject_StubWithCallback(clReleaseMemObject_testgetObjectInfo);
+    cl::BufferGL buffer(contextPool[0], flags, bufobj, &err);
+
+    buffer.getObjectInfo(&type, &bufobj);
```
bashbaug, line 4383: Couple of suggestions to improve coverage here:

1. Please test that the return value for `buffer.getObjectInfo()` is `CL_SUCCESS`, as expected.
2. Please test that `type` and `bufobj` are returned as expected.
3. In order to propertly test (2) we should initialize the values of `type` and `bufobj` to something different, because otherwise this test will pass if this query does nothing.

tests/test_openclhpp.cpp
```diff
@@ -4336,4 +4337,52 @@ void testTemplateGetImageRequirementsInfo()
...
+        *errcode_ret = CL_SUCCESS;
+    return make_mem(0);
+}
+
+static cl_int clGetGLObjectInfo_testgetObjectInfo(cl_mem memobj,
+                                                  cl_gl_object_type *type,
+                                                  cl_GLuint *gl_object_name,
+                                                  int num)
+{
+    memobj = make_mem(0);
```
bashbaug, line 4360: `memobj` is an input, so we should be testing here that it is equal to `make_mem(0)`, not assigning to it.

bashbaug (commented):
Thanks, this is building for me now.

Here are some suggestions to test this functionality more thoroughly.

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Test added

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Checks moved

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Changed to have 2 iterations

tests/test_openclhpp.cpp
bashbaug: Thanks, I don't think we need the `switch (num_calls)` though, since we're only creating one `cl::BufferGL`.

To be clear, I think only creating one `cl::BufferGL` is fine, so the fix is to remove the `switch (num_calls)`, not to create another `cl::BufferGL`.

tests/test_openclhpp.cpp
```diff
@@ -4336,4 +4337,83 @@ void testTemplateGetImageRequirementsInfo()
...
+void testgetObjectInfo() {
+    cl_mem_flags flags = 0;
+    cl_int err = 0;
+    cl_GLuint bufobj = 0;
+    cl_mem memobj = make_mem(0);
+    cl_gl_object_type type = CL_GL_OBJECT_TEXTURE2D_ARRAY;
+    clGetGLObjectInfo_StubWithCallback(clGetGLObjectInfo_testgetObjectInfo);
+    clCreateFromGLBuffer_StubWithCallback(
+        clCreateFromGLBuffer_testgetObjectInfo);
+    clReleaseMemObject_StubWithCallback(clReleaseMemObject_testgetObjectInfo);
```
bashbaug, line 4403: If you use this then you won't need to define a specific stub for `clReleaseMemObject`:

```suggestion
    clReleaseMemObject_ExpectAndReturn(make_mem(0), CL_SUCCESS);
```

tests/test_openclhpp.cpp
```diff
@@ -4336,4 +4337,83 @@ void testTemplateGetImageRequirementsInfo()
...
+    TEST_ASSERT_EQUAL(memobj, make_mem(0));
+    switch (num) {
+    case 0:
+        TEST_ASSERT_EQUAL(CL_GL_OBJECT_TEXTURE2D_ARRAY, *type);
+        *type = CL_GL_OBJECT_BUFFER;
+        break;
+    case 1:
+        TEST_ASSERT_EQUAL(CL_GL_OBJECT_BUFFER, *type);
+        *type = CL_GL_OBJECT_TEXTURE2D_ARRAY;
+        ret = CL_DEVICE_NOT_FOUND;
```
bashbaug, line 4379: There are a couple of problems here:

1. The `type` is an output value, so we really don't care what gets passed in for it.  This check should be on the caller, not in this stub.
2. We shouldn't be returning an error from this function because it causes the executables with exceptions enabled to throw an exception, which causes the test to fail.  Unless an error condition really is occurring for some reason - and it isn't in any of these tests - we shoudl return `CL_SUCCESS`.

tests/test_openclhpp.cpp
bashbaug: Thanks, I don't think the two iterations are necessary though, especially since we are only creating one `cl::BufferGL`.

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Removed switch

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Remove iteration

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Fixed

bashbaug (approved):
LGTM - thanks!

PR #235, opened 2023-05-31
test for getPlatformVersion

Bogumil-Sapinski-Mobica:


tests/test_openclhpp.cpp
```diff
@@ -3407,6 +3407,49 @@ void testLinkProgramWithVectorProgramInput(void)
...
 }
 
+void testGetPlatformVersion_1_1(void)
+{
+    cl_platform_id platform = make_platform_id(0);
+
+    clGetPlatformInfo_StubWithCallback(clGetPlatformInfo_version_1_1);
+
+    cl_uint version = cl::detail::getPlatformVersion(platform);
+    TEST_ASSERT_EQUAL_UINT32(0b10000000000000001, version);
```
bashbaug, line 3417: Nitpick: in the `opencl.hpp` code we check for hex constants instead, so it'd be clearer to use them here vs. binary constants.  Same goes for the other new tests, also.

```suggestion
    TEST_ASSERT_EQUAL_UINT32(0x10001, version);
```

bashbaug (commented):
This all looks correct, but we haven't had unit tests for any of the `cl::detail` functions previously.

Do we want to test at this level, or would it be better to have unit tests for the parts of the code that make use of details like these instead?

tests/test_openclhpp.cpp
Bogumil-Sapinski-Mobica: Changed to hexadecimal.

pwisniewskimobica:
> This all looks correct, but we haven't had unit tests for any of the `cl::detail` functions previously.
> 
> Do we want to test at this level, or would it be better to have unit tests for the parts of the code that make use of details like these instead?

@bashbaug thanks for reply. That function is not covered by tests so I think its worth to stay like that.

Bogumil-Sapinski-Mobica:
Hi @bashbaug,
I understand your point  regarding `cl::detail`. Let me explain why I think such tests are important.
It is possible to test the result of getPlatformVersion. However, these tests will be indirect in terms of the result and how the method under test is invoked.
Personally, I think there should be a clearer separation between API and library internals. The API should have a test coverage of at least 80%. Internals, it would be good to test the most popular methods and the critical ones.
getPlatformVersion could be such a method. Its result is used for inference in many important methods.

bashbaug (approved):

PR #237, opened 2023-06-13
UT for clcreateSubDevices

piotr-wozniak-mobica:


tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,86 @@ void testDevice_GetInfo_CLDeviceName()
...
+  const cl_device_partition_property_ext properties =
+      CL_DEVICE_PARTITION_TYPES_EXT;
+  std::vector<cl::Device> devices;
+  cl_platform_id platform = make_platform_id(0);
+
+  clGetPlatformInfo_StubWithCallback(
+      clGetPlatformInfo_testDevice_createSubDevices);
+  clGetDeviceInfo_StubWithCallback(clGetDeviceInfo_testDevice_createSubDevices);
+  clCreateSubDevices_StubWithCallback(
+      clCreateSubDevices_testDevice_createSubDevices);
```
bashbaug, line 3700: I think this all can be replaced by:

```suggestion
    clGetDeviceInfo_StubWithCallback(clGetDeviceInfo_platform);
    clGetPlatformInfo_StubWithCallback(clGetPlatformInfo_version_1_1);
```

Then, many of the stub functions defined above are no longer needed (all but `clCreateSubDevicesEXT_testDevice_createSubDevices`).

tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,86 @@ void testDevice_GetInfo_CLDeviceName()
...
+
+  clCreateSubDevicesEXT_StubWithCallback(
+      clCreateSubDevicesEXT_testDevice_createSubDevices);
+
+  cl_int ret = devicePool[0].createSubDevices(&properties, &devices);
+  TEST_ASSERT_EQUAL(CL_SUCCESS, ret);
+  ret = devicePool[1].createSubDevices(&properties, &devices);
+  TEST_ASSERT_EQUAL(CL_DEVICE_NOT_FOUND, ret);
+  ret = devicePool[2].createSubDevices(&properties, &devices);
+  TEST_ASSERT_EQUAL(CL_DEVICE_NOT_FOUND, ret);
```
bashbaug, line 3710: These tests that return error codes will all "fail" for the `CL_HPP_ENABLE_EXCEPTIONS` tests, because the error conditions will cause exceptions to be thrown.  Options:

1. They should only be run with `CL_HPP_ENABLE_EXCEPTIONS` is not defined.
2. They should be rewritten so they do not return error codes.

tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,86 @@ void testDevice_GetInfo_CLDeviceName()
...
+    int num) {
+    return 0;
+}
+
+static cl_int clCreateSubDevicesEXT_testDevice_createSubDevices(
+    cl_device_id device_in, const cl_device_partition_property_ext *properties,
+    cl_uint n, cl_device_id *out_devices, cl_uint *num, int num_calls) {
+  cl_int ret = CL_SUCCESS;
+  if (device_in == make_device_id(0)) {
+    return CL_SUCCESS;
```
bashbaug, line 3682: Can we test a few more conditions here (and then in the tester below, also)?  Examples:

1. Test that the partition properties are passed through properly.
2. Write output subdevices to the `out_devices` array and verify that they are returned properly.

bashbaug (commented):
Here are a few comments and suggested improvements.  The issue where exceptions are being thrown needs to be resolved before merging.

As a meta comment, it would be great to test the core OpenCL 1.2 sub-device functionality in addition to the `cl_ext_device_fission` extension!

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Changed

tests/test_openclhpp.cpp
piotr-wozniak-mobica: We need some "fails" to cover all options. CL_HPP_ENABLE_EXCEPTIONS added

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Changed as suggested

tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,63 @@ void testDevice_GetInfo_CLDeviceName()
...
-} 
+}
+
+#if defined(cl_ext_device_fission)
+static cl_int clCreateSubDevices_testDevice_createSubDevices(
+    cl_device_id in_device, const cl_device_partition_property *properties,
+    cl_uint num_devices, cl_device_id *out_devices, cl_uint *num_devices_ret,
+    int num) {
+    return 0;
+}
```
bashbaug, line 3641: This entire function looks like dead code.  If this is not used can it be removed?

tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,63 @@ void testDevice_GetInfo_CLDeviceName()
...
+
+static cl_int clCreateSubDevicesEXT_testDevice_createSubDevices(
+    cl_device_id device_in, const cl_device_partition_property_ext *properties,
+    cl_uint n, cl_device_id *out_devices, cl_uint *num, int num_calls) {
+  cl_int ret = CL_SUCCESS;
+
+  TEST_ASSERT_EQUAL(CL_DEVICE_PARTITION_TYPES_EXT, *properties);
```
bashbaug, line 3648: This is OK for a unit test but do note that `CL_DEVICE_PARTITION_TYPES_EXT` is not a valid partition property according to the cl_ext_device_fission [spec](https://registry.khronos.org/OpenCL/extensions/ext/cl_ext_device_fission.txt).  If we wanted this to be a more real-world test case then we should choose different partition properties, such as perhaps `CL_DEVICE_PARTITION_EQUALLY_EXT` or `CL_DEVICE_PARTITION_BY_AFFINITY_DOMAIN_EXT`.

tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,63 @@ void testDevice_GetInfo_CLDeviceName()
...
+  if(nullptr != out_devices){
+    out_devices[0] = make_device_id(0);
+  }
+  else{
+    *num = 1;
+  }
```
bashbaug, line 3654: These aren't exactly real-world return values, either:

1. The number of output sub-devices should always be written to if it is non-nullptr, not only if out_devices is nullptr.
2. I don't think it's possible to partition a device into a single sub-device.

For a unit test I suppose this is OK, though it would be better to have more real-world values instead.

tests/test_openclhpp.cpp
```diff
@@ -3626,7 +3630,63 @@ void testDevice_GetInfo_CLDeviceName()
...
+  if (device_in == make_device_id(0)) {
+    return CL_SUCCESS;
+  } else if (device_in == make_device_id(1) && nullptr == num) {
+    return CL_DEVICE_NOT_FOUND;
```
bashbaug, line 3658: 1. Why does it matter if `nullptr == num`?
2. `CL_DEVICE_NOT_FOUND` is not a valid error code for this function.  Could we pick a valid error code instead, maybe `CL_INVALID_DEVICE` or `CL_INVALID_VALUE`?

bashbaug (commented):
Thanks, this is OK as a unit test but I'd prefer to see more real-world parameters and return values if possible.  Can you please take a look at the comments below?

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Sure, removed.

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Changed to proper value.

tests/test_openclhpp.cpp
piotr-wozniak-mobica: Changed to always write number of sub-devices

tests/test_openclhpp.cpp
piotr-wozniak-mobica: 1. I wanted a function to fail in right place but it can be without this check also
2. Chenged to valid error code

bashbaug (approved):
There is more that we could do here, but this is an improvement, so no reason to hold off merging it.

PR #242, opened 2023-06-23
fixed warning unused function in UT

Bogumil-Sapinski-Mobica:


tests/test_openclhpp.cpp
```diff
@@ -1334,7 +1335,6 @@ static cl_mem clCreateBufferWithProperties_testBufferWithProperties(
 
 void testBufferWithProperties(void)
```
alycm, line 1336: Looks like this function is used somewhere not compatible with this `#if` guard change.

bashbaug (approved):

PR #243, opened 2023-06-23
fixed warning int-in-bool-context in UT

Bogumil-Sapinski-Mobica:


tests/test_openclhpp.cpp
```diff
@@ -1200,7 +1200,7 @@ static cl_mem clCreateBuffer_testBufferConstructorContextIterator(
     (void) num_calls;
 
     TEST_ASSERT_EQUAL_PTR(make_context(0), context);
-    TEST_ASSERT_BITS(CL_MEM_COPY_HOST_PTR, flags, !CL_MEM_COPY_HOST_PTR);
+    TEST_ASSERT_BITS(CL_MEM_COPY_HOST_PTR, flags, ~CL_MEM_COPY_HOST_PTR);
     TEST_ASSERT_BITS(CL_MEM_READ_ONLY, flags, CL_MEM_READ_ONLY);
```
bashbaug, line 1204: I'm OK with this fix, but would it be better to test that flags is exactly the way we expect it to be (in other words, that it equals `CL_MEM_READ_ONLY`), rather than testing these two specific bits?

Bogumil-Sapinski-Mobica: I do not understand a bit. I have just changed the `operator !()` which returns boolean and this is not what we need in here. I did not change the logic. And all that we want in this test of the ctor is to verify if flags were initialized to 0 - specifically interesting us bits. Thus negation of the mask serves as expected output.
Do you mean that assert from line 1204 is enough? Should I remove this check?

bashbaug: Sorry about that, let me try to explain better.  It seems a little odd that we are testing that the flags bitfield a) does NOT contain CL_MEM_COPY_HOST_PTR, and b) does contain CL_MEM_READ_ONLY.  Do we ever think that any other bits might be included?  If not, can we just test for the specific value we expect?

```suggestion
    TEST_ASSERT_EQUAL(flags, CL_MEM_READ_ONLY);
```

Bogumil-Sapinski-Mobica: Suggestions applied.

bashbaug (approved):
Thanks, latest version LGTM.

Issue #233, opened 2023-05-11
Naming Inconsistency for Extension Objects


bashbaug:
We currently have a naming inconsistency for command buffer and semaphore extension objects:

```c
#ifdef cl_khr_semaphore
    class Semaphore;
#endif
#if defined(cl_khr_command_buffer)
    class CommandBufferKhr;
    class MutableCommandKhr;
#endif // cl_khr_command_buffer
```

Note the `Khr` suffix on `CommandBuffer` and `MutableCommand` objects but not on `Semaphore` objects.

Do we want a `Khr` suffix on these objects, at least while they are extension objects?

If we want some sort of decoration while these are extension objects is a suffix the best mechanism?  We could consider a namespace instead, for example `cl::khr::CommandBuffer`.

aharon-abramson:
Yes, I prefer the namespace idea.

Bogumil-Sapinski-Mobica:
!assign

bashbaug:
Discussed in the July 11th special teleconference.  We prefer the namespace idea, if this will work.

We should give thought how the C++ bindings would change if any of these objects became core features sometime in the future.

PR #244, opened 2023-07-14
Fix naming inconsistency khr

Bogumil-Sapinski-Mobica:


bashbaug (commented):
Thanks, this looks like the right direction, but the inconsistency still exists because now the `CommandBuffer` and `MutableCommandBuffer` classes are in the `khr` namespace, but the `Semaphore` class is not.  Are you planning to move the `Semaphore` class into the `khr` namespace also?

As an aside, it looks like parts of this file were reformatted but had no other changes.  I would generally advise against this, but if would like to reformat parts of the file could you please do it as the first or last commit?  This will make it easier to see and review the non-formatting changes - thanks!

Bogumil-Sapinski-Mobica:
Hi @bashbaug 
sorry for mixing formatting changes with task related changes. 
Would you like me to create new PR with these changes or should I continue with Semaphore?

bashbaug:
Related to https://github.com/KhronosGroup/OpenCL-CLHPP/issues/233

tests/test_openclhpp.cpp
```diff
@@ -516,8 +512,9 @@ void testMoveAssignContextNonNull(void);
 void testMoveAssignContextNull(void);
 void testMoveConstructContextNonNull(void);
 void testMoveConstructContextNull(void);
+namespace cl{
 MAKE_MOVE_TESTS(Context, make_context, clReleaseContext, contextPool)
-
+}
```
bashbaug, line 517: This is a little ugly.  How about this as an alternative:

1. Separate out the "type" and the "typestr", whereas we just have a single "type" currently.
2. Use `Context` as the "typestr" in this case, and e.g. `CommandBuffer` as the "typestr" in the command buffer cases.
3. Use `cl::Context` as the "type" in this case, and e.g. `cl::khr::CommandBuffer` as the "type" in the command buffer cases.

I think this will require fewer diffs, and it won't require putting the tests themselves inside of the `cl` or `cl::khr` namespaces, which they really shouldn't be doing.

bashbaug (commented):
Thanks, this mostly looks good to me, thanks for making these changes.

I have one specific comment for the test changes, see below.

Additionally, I'd like to reiterate that formatting related changes make it much harder to review this PR than it should be.  I'd strongly recommend reverting the formatting changes to minimize diffs.

Finally it appears that this PR needs a rebase since there are merge conflicts.

tests/test_openclhpp.cpp
kamil-goras-mobica: @bashbaug In last commit (73c5491 https://github.com/KhronosGroup/OpenCL-CLHPP/pull/244/commits/73c5491d2497140d2c61352c49edccfd0e9dfa00) I have added corrections and removed not needed style corrections.

bashbaug (approved):
Apologies for the long review delay.  The updated change LGTM.  There needs to be a rebase (in the test code?) now, but hopefully that is straightforward.

Noting for completeness: these changes do break backward compatibility so they will require code changes for applications using the C++ bindings with semaphores and command buffers.

PR #245, opened 2023-07-15
fix preprocessor guards for getContextPlatformVersion

### LINKED ISSUE #108: error: ‘getContextPlatformVersion’ is not a member of ‘cl::detail’ in some cases
I got the error `error: ‘getContextPlatformVersion’ is not a member of ‘cl::detail’` when building my project:

It seems that on line [7078](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L7075-L7078) (and [7144](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L7141-L7144), [7209](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L7206-L7209), [7276](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L7273-L7276), [7327](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L7324-L7327) and [7378](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L7375-L7378)) `getContextPlatformVersion` is used in a `#if`-block that is enabled if `CL_HPP_MINIMUM_OPENCL_VERSION`  is less than 200:
```
#if CL_HPP_TARGET_OPENCL_VERSION >= 200 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
            // Run-time decision based on the actual platform
            {
                cl_uint version = detail::getContextPlatformVersion(context());
```

However, the (only) definition of `getContextPlatformVersion` (on line [1738](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L1738)) is only active if `CL_HPP_MINIMUM_OPENCL_VERSION` is less than 120 (checked at line [1700](https://github.com/KhronosGroup/OpenCL-CLHPP/blob/master/include/CL/cl2.hpp#L1700)):
```
#if CL_HPP_TARGET_OPENCL_VERSION >= 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 120
```
The error should therefore always ocure if `CL_HPP_TARGET_OPENCL_VERSION` is 200 and `CL_HPP_MINIMUM_OPENCL_VERSION` is 120.

I could complile my project by changing that line (1700) to this:
```
#if CL_HPP_TARGET_OPENCL_VERSION >= 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
```
But I don't know if that is the proper solution or if that would have other side effects, therefore I didn't create a pull request and instead created this issue...

bashbaug:
fixes #108 

This change corrects the preprocessor guards around detail functions to get the OpenCL version from a platform, device, or context.

* The preprocessor guard for `CL_HPP_TARGET_OPENCL_VERSION` needs to be the **smallest** value that callers of these functions test for.  This is correctly `120` for current usages of these functions.
* The preprocessor guard for `CL_HPP_MINIMUM_OPENCL_VERSION` needs to be the **largest** value that callers of these functions test for.  This needs to be `200`, for example for command queue properties.

Debatably the preprocessor guards around these functions should be removed completely, but at least this way things won't be broken while we figure out if this is the right thing to do or not.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -2086,7 +2086,7 @@ struct ReferenceHandler<cl_mutable_command_khr>
 #endif // cl_khr_command_buffer
 
 
-#if CL_HPP_TARGET_OPENCL_VERSION >= 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 120
+#if CL_HPP_TARGET_OPENCL_VERSION >= 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
 // Extracts version number with major in the upper 16 bits, minor in the lower 16
 static cl_uint getVersion(const vector<char> &versionInfo)
 {
@@ -2136,7 +2136,7 @@ static cl_uint getContextPlatformVersion(cl_context context)
     clGetContextInfo(context, CL_CONTEXT_DEVICES, size, devices.data(), nullptr);
     return getDevicePlatformVersion(devices[0]);
 }
-#endif // CL_HPP_TARGET_OPENCL_VERSION >= 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 120
+#endif // CL_HPP_TARGET_OPENCL_VERSION >= 120 && CL_HPP_MINIMUM_OPENCL_VERSION < 200
 
 template <typename T>
 class Wrapper


bashbaug:
Need to file an issue (or add a test) to ensure these types of failures are caught in the future.

bashbaug:
I've added a test to ensure these issues will be caught in the future - see https://github.com/KhronosGroup/OpenCL-CLHPP/pull/252.  These tests fail currenlty, but will pass once this PR is merged.

Kerilk (approved):

bashbaug:
Merging as discussed in the December 5th teleconference.

PR #252, opened 2023-09-17
add tests for all valid combinations of min and target version

bashbaug:
This PR adds very simple tests to ensure there are no compile errors when compiling the C++ headers for all combinations of the minimum and target OpenCL versions.  Some of these tests will fail until #245 is merged.


Kerilk (approved):
LGTM.

PR #253, opened 2023-09-19
UT for getHostTimer

piotr-wozniak-mobica:


tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,25 @@ void testgetObjectInfo() {
...
     TEST_ASSERT_EQUAL(bufobj, 0);
 }
+#if CL_HPP_TARGET_OPENCL_VERSION >= 210
+static cl_int clGetHostTimer_testgetHostTimer(cl_device_id device,
+                                              cl_ulong *host_timestamp,
+                                              int num_calls) {
+    TEST_ASSERT_EQUAL_PTR(devicePool[0](), device);
+    TEST_ASSERT_EQUAL(0, num_calls);
+    *host_timestamp = 1;
+    return 0;
```
bashbaug, line 4527: Note for later: it would be clearer to `return CL_SUCCESS`.

tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,25 @@ void testgetObjectInfo() {
...
+}
+
+void testgetHostTimer() {
+    cl_ulong retVal = 0;
+    cl_int *error = nullptr;
+
+    clGetHostTimer_StubWithCallback(clGetHostTimer_testgetHostTimer);
+    retVal = devicePool[0].getHostTimer(error);
```
bashbaug, line 4535: Note for later: if we're going to pass in an error code it would be good to verify it's coming back as `CL_SUCCESS`.

Or, we could just remove the error code pointer, since it is `nullptr` by default.

bashbaug (approved):
There is more that we could do here, but this is an improvement, so no reason to hold off merging it.  Thanks!

PR #254, opened 2023-09-19
UT for createSubBuffer

piotr-wozniak-mobica:


tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,37 @@ void testgetObjectInfo() {
...
+    const void *buffer_create_info = nullptr;
+    cl_int *err = nullptr;
+    static cl::Buffer ret;
+
+    clCreateSubBuffer_StubWithCallback(clCreateSubBuffer_testCreateSubBuffer);
+    ret = bufferPool[0].createSubBuffer(flags, buffer_create_type,
+                                        buffer_create_info, err);
+
+    TEST_ASSERT_EQUAL_PTR(make_mem(1), ret());
+    ret() = nullptr;
```
bashbaug, line 4547: Please make this variable non-static:

```suggestion
    clCreateSubBuffer_StubWithCallback(clCreateSubBuffer_testCreateSubBuffer);
    clReleaseMemObject_ExpectAndReturn(make_mem(1), CL_SUCCESS);
    cl::Buffer ret = bufferPool[0].createSubBuffer(flags, buffer_create_type,
                                        buffer_create_info, err);

    TEST_ASSERT_EQUAL_PTR(make_mem(1), ret());
```

tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,37 @@ void testgetObjectInfo() {
...
 }
+#if CL_HPP_TARGET_OPENCL_VERSION >= 110
+static cl_mem clCreateSubBuffer_testCreateSubBuffer(
+    cl_mem buffer, cl_mem_flags flags, cl_buffer_create_type buffer_create_type,
+    const void *buffer_create_info, cl_int *errcode_ret, int num_calls) {
+    (void)errcode_ret;
+    TEST_ASSERT_EQUAL(make_mem(0), buffer);
+    TEST_ASSERT_EQUAL(0, flags);
+    TEST_ASSERT_EQUAL(0, buffer_create_type);
+    TEST_ASSERT_EQUAL_PTR(nullptr, buffer_create_info);
```
bashbaug, line 4528: We really ought to be testing more real-world values here:

* `buffer` - this one is OK.
* `flags` - a value of `0` is legal, but it's not a very useful value to test to see if the flags are being passed through properly.  Maybe pass something like `CL_MEM_READ_ONLY` instead?
* `buffer_create_type` - this should be `CL_BUFFER_CREATE_TYPE_REGION`, since this is the only buffer creation type described in the spec.
* `buffer_create_info` - this should be a non-NULL pointer to a `cl_buffer_region` structure.  Ideally, we should check that the pointer is non-NULL, and that the contents of the structure are what we expect.

bashbaug (changes requested):

PR #255, opened 2023-09-19
Fix throwing from SVMAllocator with exceptions disabled

dendy:


bashbaug (approved):
Nice catch!

It'd be great to have a test for this scenario.

dendy:
Thanks for the review. Do we need to add someone specifically to proceed with merging the patch?

bashbaug:
> Thanks for the review. Do we need to add someone specifically to proceed with merging the patch?

Nah, I usually wait a bit just in case anyone else has review comments, but this is a pretty straightforward fix.  I'll just merge this now.   (It would be nice to have a test, but this can always be done later...)

PR #256, opened 2023-09-19
use the new semaphore specific device handle list enum

bashbaug:
Updates the C++ bindings to use the new semaphore-specific device handle list enum rather than the older overloaded enum.

Includes a preprocessor check so these changes will work with both older and newer headers.  When newer headers become widespread we can remove this check - this could be part of this PR or in a subsequent PR.

For reference, the PR to update the headers with the new enums is: https://github.com/KhronosGroup/OpenCL-Headers/pull/240


bashbaug:
I've updated the handling of the new enum, as we discussed:

- Older headers are still supported, but the older non-semaphore-specific device handle list enum is no longer supported.
- To query the device handle list for a semaphore applications will need to update headers or otherwise use the new semaphore-specific device handle list enum.

Kerilk (approved):
LGTM!

bashbaug:
Merging as discussed in the November 14th teleconference.

PR #257, opened 2023-09-22
UT for enqueueMarkerWithWaitList

piotr-wozniak-mobica:


tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,35 @@ void testgetObjectInfo() {
...
+    if (event != nullptr) {
+            *event = make_event(1);
+    }
+    return CL_SUCCESS;
+}
+
+void testenqueueMarkerWithWaitList() {
+    cl_int ret = CL_DEVICE_NOT_FOUND;
+    cl::Event event;
+    cl::vector<cl::Event> events = {event};
```
bashbaug, line 4537: This is not valid "real-world" code - the set of events in the event wait list must be different than the event that is signaled.  Coudl you please make a specific event to wait upon instead?  This would also allow us to ensure that it was passed through correctly.

Something like:

```suggestion
    cl::Event event;
    cl::vector<cl::Event> events = {cl::Event(make_event(0))};
```

bashbaug (changes requested):

kamil-goras-mobica:
@bashbaug correction added.

PR #259, opened 2023-09-22
UT for EnqueueMapBuffer

piotr-wozniak-mobica:


tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,41 @@ void testgetObjectInfo() {
...
+    cl_event *event, cl_int *errcode_ret, int num_calls) {
+    (void)command_queue;
+    TEST_ASSERT_EQUAL_PTR(make_mem(0), buffer);
+    TEST_ASSERT_EQUAL(CL_TRUE, blocking_map);
+    TEST_ASSERT_EQUAL(CL_MAP_WRITE, map_flags);
+    TEST_ASSERT_EQUAL(0, offset);
+    TEST_ASSERT_EQUAL(sizeof(int) * 1024, size);
+    TEST_ASSERT_EQUAL(0, num_events_in_wait_list);
+    TEST_ASSERT_EQUAL_PTR(nullptr, event_wait_list);
+    TEST_ASSERT_NOT_NULL(event);
```
bashbaug, line 4533: The event pointer is an optional output, so if the pointer is non-NULL then we should be writing a value to it (e.g. `make_event(N)`, and checking that the expected value was written in the calling function.

See, for example:

* https://github.com/KhronosGroup/OpenCL-CLHPP/blob/f96b8f865682e0bc27f6f61c34c5e75f76490ee4/tests/test_openclhpp.cpp#L3772-L3774
* https://github.com/KhronosGroup/OpenCL-CLHPP/blob/f96b8f865682e0bc27f6f61c34c5e75f76490ee4/tests/test_openclhpp.cpp#L3791

tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,41 @@ void testgetObjectInfo() {
...
+    cl::Event event_data(make_event(0), false);
+    cl::Event *event = &event_data;
+    const cl::vector<cl::Event> *events = nullptr;
+    cl_int *err = nullptr;
+    void *ret = nullptr;
+
+    clEnqueueMapBuffer_StubWithCallback(
+        clEnqueueMapBuffer_testenqueueMapBuffer);
+    ret = enqueueMapBuffer(bufferPool[0], blocking, flags, offset, size, events,
+                           event, err);
```
bashbaug, line 4552: I don't have a very strong opinion about this, but it would require less code to pass in some of these values directly vs. assigning them to a variable first.

```suggestion
    ret = enqueueMapBuffer(bufferPool[0], CL_TRUE, CL_MAP_WRITE, 0, sizeof(int) * 1024, events,
                           event, err);
```

kamil-goras-mobica:
@bashbaug Added corrections.

PR #260, opened 2023-09-22
UT for enqueueReleaseGLObjects

piotr-wozniak-mobica:


tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,38 @@ void testgetObjectInfo() {
...
+            *event = make_event(1);
+    }
+    return CL_SUCCESS;
+}
+
+void testenqueueReleaseGLObjects() {
+    cl::Memory obj;
+    cl::vector<cl::Memory> mem_objects = {obj};
+    cl::Event event;
+    cl::vector<cl::Event> events = {event};
```
bashbaug, line 4540: See related comment: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/257#discussion_r1394908128

tests/test_openclhpp.cpp
```diff
@@ -4517,4 +4517,38 @@ void testgetObjectInfo() {
...
+    TEST_ASSERT_EQUAL(0, num_calls);
+    if (event != nullptr) {
+            *event = make_event(1);
+    }
+    return CL_SUCCESS;
+}
+
+void testenqueueReleaseGLObjects() {
+    cl::Memory obj;
+    cl::vector<cl::Memory> mem_objects = {obj};
```
bashbaug, line 4538: `mem_objects` (and hence `obj`) are input values to the function we are testing, so we need to ensure they are passed through correctly.  We should be able to use an object from the bufferPool for it.

kamil-goras-mobica:
@bashbaug Corrections added.

PR #262, opened 2023-09-25
Updated CI, added CD

mfep:
- Added .gitignore file,
- Added .clang-format (based on OpenCL-SDK's corresponding file) and format enforcing script.
- Minor changes to source files to avoid warnings on all compilers tested in CI.
  - `-Werror`/`/WX` is enabled in CI
- Refactored the `CMakeLists.txt` to be able to produce a binary Debian package using `cpack`.
  - CMake minimum version is bumped to 3.16.
- Updated the CI (pre-release) script.
  - Removed the previous separate workflows for Windows, MacOS and Linux. All CI scripts reside in the `presubmit` workflow now.
  - Added a job that checks the changed code for formatting conformance via `git-clang-format`.
  - Updated Linux CI to use a new container image
    - Most jobs in the matrix are based on Ubuntu 22.04
    - Compilers gcc-11, gcc-13, clang-14, clang-16 and gcc-9 on Ubuntu 20.04
    - Binary targets: x86-64 and x86 on Ubuntu 20.04
    - C++ standards: 11, 17
    - CMake versions: 3.26 and 3.16 on Ubuntu 20.04
    - CMake generators: Ninja Multi-Config, Unix Makefiles
    - Added the generation of the Debian package, testing the package, and testing the install tree to the workflow
  - Updated Windows CI, which runs on Windows Server 2022
    - MSVC toolsets: v141, v142, v143, clang-cl (this latter with msbuild only)
    - Binary targets: x86-64, x86
    - C++ standards: 11, 17
    - CMake: system installed, currently 3.27
    - CMake generators: Ninja Multi-Config, Visual Studio 17 2022
  - Updated MacOS CI, which runs on MacOS 13
    - Compilers: Apple-Clang 14, gcc-11, gcc-13 (gcc with Ninja Multi-Config only)
    - Binary target: x86-64
    - C++ standards: 11, 17
    - CMake: system installed, currently 3.27
    - CMake generators: Ninja Multi-Config, XCode
  - Added Android CI, which is build only (virtualization is not available with the default GitHub runners)
    - Binary targets: x86-64, arm64
    - Android API levels: 19, 33
- Added a release workflow which triggers on the push of a `v*` tag
  - Added `DebSourcePkg.cmake` which is intended to run in CMake script mode. This script generates the `debian/control`, `debian/changelog` and ``debian/rules` files which are required to build a Debian source package.
  - The generated source package produces 2 binary packages: one containing the CLHPP headers (opencl-clhpp-headers), and another that contains the Doxygen-generated documentation (opencl-clhpp-headers-doc).
  - The release workflow generates the Debian source package and pushes it to a prescribed Launchpad PPA. The details of the release (e.g. maintainer, signing key) has to be set up as repository variables.
  - Detailed instruction regarding the release procedure is going to be added to the OpenCL SDK repository.
  
This PR is considered to be complete, albeit review remarks and/or changes to related PRs might warrant minor updates.


.github/workflows/presubmit.yml
```diff
@@ -0,0 +1,876 @@
...
+        CXXSTD: [11, 17]
+        CONF:
+          - GEN: Unix Makefiles
+            CONFIG: Debug
+          - GEN: Unix Makefiles
+            CONFIG: Release
+          - GEN: Ninja Multi-Config
+            CONFIG: Release
+        IMAGE:
+          - streamhpc/opencl-sdk-base:ubuntu-22.04-20230717
```
bashbaug, line 56: Checking: should this still be using a streamhpc image?

scripts/check-format.sh
```diff
@@ -0,0 +1,55 @@
...
+
+# Check for no-ops
+grep '^no modified files to format$\|^clang-format did not modify any files$' \
+    "$scratch" > /dev/null && exit 0
+
+# Dump formatting diff and signal failure
+printf \
+"\033[31m==== FORMATTING VIOLATIONS DETECTED ====\033[0m
+run '\033[33m%s --style=file %s %s\033[0m' to apply these formating changes\n\n" \
+"$GIT_CLANG_FORMAT" "$*" "$SOURCE_COMMIT"
```
bashbaug, line 52: Does the current file pass these formatting checks?  Or is the intent that the formatting requirements will apply to any new code?

bashbaug (approved):
LGTM, just a few minor questions, thanks!

.github/workflows/presubmit.yml
mfep: Yes, we think that this is not a blocker for merge (the already merged Headers CI also uses these images). Is this correct?

The corresponding Dockerfiles are provided as a PR: https://github.com/KhronosGroup/DockerContainers/pull/32

scripts/check-format.sh
mfep: No, the formatting check is applied to the new code only. The base for the diff is the PR target branch in the case of a PR, or `main`, in the case of a push.

bashbaug:
Merging as discussed in the June 25th teleconference.

PR #276, opened 2023-10-26
Ut for update mutable commands

kamil-goras-mobica:


bashbaug:
Fixed by #308 - closing.

PR #278, opened 2023-10-30
Added wrapper constructors for clCreateImageWithProperties

shajder:
Added new constructors for `Image1D`, `Image1DArray`, `Image1DBuffer`, `Image2D`, `Image2DArray`, `Image3D` to wrap `clCreateImageWithProperties` features. 

I have tested/discussed most of new constructors in real environment, [here](https://github.com/Mobica/OpenCL-SDK/tree/test-clcreateimagewithproperties) is modified `juliavk` sample (samples/vulkan/00_juliavk) adapted from @bashbaug [repository](https://github.com/bashbaug/SimpleOpenCLSamples) to test new constructors.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -4832,6 +4832,42 @@ public:
     //! \brief Default constructor - initializes to nullptr.
     Image1D() { }
 
+#if CL_HPP_TARGET_OPENCL_VERSION >= 300
+    /*! \brief Constructs a Image1D with specified properties.
+     *
+     *  Wraps clCreateImageWithProperties().
+     *
+     *  \param properties Optional list of properties for the image object and
+     *                    their corresponding values. The non-empty list must
+     *                    end with 0.
+     *  \param host_ptr Storage to be used if the CL_MEM_USE_HOST_PTR flag was
+     *                  specified. Note alignment & exclusivity requirements.
+     */
+    Image1D(const Context &context, const vector<cl_mem_properties> &properties,
+            cl_mem_flags flags, ImageFormat format, size_type width,
+            void *host_ptr = nullptr, cl_int *err = nullptr) {
+      cl_int error;
+
+      cl_image_desc desc = {};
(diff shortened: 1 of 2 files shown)


bashbaug (approved):
LGTM, this seems like a useful addition.

```
     *  \param properties Optional list of properties for the image object and
     *                    their corresponding values. The non-empty list must
     *                    end with 0.
```

Is it desirable to detect cases when the list of properties does NOT end with zero and return some kind of error?

shajder:
> LGTM, this seems like a useful addition.
> 
> ```
>      *  \param properties Optional list of properties for the image object and
>      *                    their corresponding values. The non-empty list must
>      *                    end with 0.
> ```
> 
> Is it desirable to detect cases when the list of properties does NOT end with zero and return some kind of error?

Sounds good for me but ... in such case we should probably do the same in other locations eg. Buffer class, right ?

bashbaug:
> Sounds good for me but ... in such case we should probably do the same in other locations eg. Buffer class, right ?

Ah, good point.  Let's save this for a later PR, then.

Last question for this PR: would it be possible to add some tests for this new functionality, similar to `testBufferWithProperties`?

shajder:
> would it be possible to add some tests for this new functionality, similar to `testBufferWithProperties`?

sure thing, are you thinking about this PR or separate ?

bashbaug:
> sure thing, are you thinking about this PR or separate ?

In this PR would be ideal, then it doesn't add to the backlog of unit test PRs we still need to review and merge.

Is this doable?  If not, we'll merge now, and add it to the list (risk seems small).

shajder:
> Is this doable?

I will complement shortly and let you know

shajder:
> would it be possible to add some tests for this new functionality

@bashbaug I just added tests for Image2D and Image3D. What is missing is the rest of Image constructors, not only new but any tests for Image1D, Image1DArray, Image2DArray, worth creating an issue ?

bashbaug:
Thanks, tests LGTM.  The CI failures are unrelated to your change, so I'll just merge this.

Issue #285, opened 2024-02-01
No Device-parametrized versions of compile()?


pjaaskel:
There are versions of build() which can limit the devices to build for, but there is only one compile() which builds for all devices in the context. Is this for some purpose that I do not see, or just not implemented yet?

bashbaug:
@pjaaskel I think #286 fixes this issue?

PR #286, opened 2024-02-02
expose header parameters in clCompileProgram

Richardk2n:
I use this to include dynamically generated headers.
Exposing this should offer more flexibility without breaking existing code due to the defaults.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -6720,7 +6720,7 @@ public:
     }
 
     cl_int compile(
-        const char* options,
+        const string options,
         const vector<Program>& inputHeaders,
         const vector<string>& headerIncludeNames,
         void (CL_CALLBACK * notifyFptr)(cl_program, void *) = nullptr,
@@ -6734,7 +6734,7 @@ public:
             object_,
             0,
             nullptr,
-            options,
+            options.c_str(),
             static_cast<cl_uint>(inputHeaders.size()),
             reinterpret_cast<const cl_program*>(inputHeaders.data()),
             reinterpret_cast<const char**>(headerIncludeNamesCStr.data()),
@@ -6744,7 +6744,7 @@ public:
     }
 
     cl_int compile(
-        const char* options,
+        const string options,
         const vector<Device>& deviceList,
         const vector<Program>& inputHeaders = vector<Program>(),
         const vector<string>& headerIncludeNames = vector<string>(),
@@ -6759,7 +6759,7 @@ public:
             object_,
             static_cast<cl_uint>(deviceList.size()),
             reinterpret_cast<const cl_device_id*>(deviceList.data()),
-            options,
+            options.c_str(),
             static_cast<cl_uint>(inputHeaders.size()),
             reinterpret_cast<const cl_program*>(inputHeaders.data()),
             reinterpret_cast<const char**>(headerIncludeNamesCStr.data()),
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -6720,14 +6720,16 @@ public:
     }
 
     cl_int compile(
-        const string options,
+        const string& options,
         const vector<Program>& inputHeaders,
         const vector<string>& headerIncludeNames,
         void (CL_CALLBACK * notifyFptr)(cl_program, void *) = nullptr,
         void* data = nullptr) const
     {
+        static_assert(sizeof(cl::Program) == sizeof(cl_program),
+            "Size of cl::Program must be equal to size of cl_program");
         vector<const char*> headerIncludeNamesCStr;
-        for(string name: headerIncludeNames) {
+        for(const string& name: headerIncludeNames) {
             headerIncludeNamesCStr.push_back(name.c_str());
         }
         cl_int error = ::clCompileProgram(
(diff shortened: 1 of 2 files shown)


include/CL/opencl.hpp
```diff
@@ -6703,6 +6703,9 @@ class Program : public detail::Wrapper<cl_program>
 #if CL_HPP_TARGET_OPENCL_VERSION >= 120
     cl_int compile(
         const char* options = nullptr,
+        cl_uint num_input_headers = 0,
+        const cl_program* input_headers = nullptr,
```
bashbaug, line 6707: Could this be a `vector` of `cl::Programs` instead?  See `linkProgram` for an example.

include/CL/opencl.hpp
```diff
@@ -6703,6 +6703,9 @@ class Program : public detail::Wrapper<cl_program>
...
+        const char** header_include_names = nullptr,
```
bashbaug, line 6708: I'm less sure about this one, but could this be a `vector` of `strings` instead?

bashbaug (commented):
Thanks, this is a nice addition!

I think this isn't quite backward compatible though, even with the default arguments.  To see why, imagine a case that called `compile()` and passed a notification function pointer (and possibly some user data) before.  This will not work after these changes because the order of the function arguments has changed.

We have several options:

1. Break compatibility.  Not ideal, but the odds of this change affecting a significant amont of code is probably low, and if any code breaks the error message and steps to fix it should be clear
2. Move the header arguments to the end of the argument list.  This will retain compatibility, although if we did this then the order of arguments will not match `clCompileProgram`.
3. Keep the current function as-is and add an overload that supports headers.  Note, the overload will need to have some non-default arguments for overload resolution to work.

I think my order of preference is (3), then (2), and then (1), though I'm open to alternatives.  What do you think?

PS: It would be great to have some unit tests for these changes!

Richardk2n:
You are right of course about compatibility. Not sure what inhibited my thinking when I wrote that.

Due to the limitation of the non default argument needing to come first and me wanting to preserve argument order, this leads to the requirement to provide options, which is ok I guess. No need to break compatibility.

While I was at it, I also wrote an overload addressing #285

Honestly, I have no clue how to write unit tests for something like this.

Question:
Should the original function and my header overload just call the second overload, that contains all parameters in order to reduce code?

bashbaug (commented):
Thanks, I think this is moving in the right direction, just a few minor things to fix.

I also think it'd be really good to have some tests to ensure this is doing what we want it to do.  Is this something you could look into adding?  If not, could you please file an issue so we don't lose track of it?  Thanks again!

include/CL/opencl.hpp
```diff
@@ -6718,6 +6718,55 @@ class Program : public detail::Wrapper<cl_program>
...
+
+    cl_int compile(
+        const string options,
+        const vector<Program>& inputHeaders,
+        const vector<string>& headerIncludeNames,
+        void (CL_CALLBACK * notifyFptr)(cl_program, void *) = nullptr,
+        void* data = nullptr) const
+    {
+        vector<const char*> headerIncludeNamesCStr;
+        for(string name: headerIncludeNames) {
```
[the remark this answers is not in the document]
Richardk2n: I prefer your recommendation.

include/CL/opencl.hpp
```diff
@@ -6718,6 +6718,55 @@ class Program : public detail::Wrapper<cl_program>
...
+            headerIncludeNamesCStr.push_back(name.c_str());
+        }
+        cl_int error = ::clCompileProgram(
+            object_,
+            static_cast<cl_uint>(deviceList.size()),
+            reinterpret_cast<const cl_device_id*>(deviceList.data()),
```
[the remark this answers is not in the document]
Richardk2n: You are right, I copied that from `cl::Program` thinking it also is just a `detail::Wrapper<cl_device_id>`
It should be fixed now.

bashbaug (approved):
Thanks, this looks really great.

I had a look at adding some tests for this feature and it's going to be complex, since compiling the program also queries the build log.  I'm fine with just the tracker issue for now.

bashbaug:
I left a PR on your repo with some tests - please take a look: https://github.com/Richardk2n/OpenCL-CLHPP/pull/1

I do wonder if we should re-think passing the options as a `string` vs. a `const char*`, for consistency and to allow passing `nullptr` options.

Richardk2n:
In my opinion using `string` is preferable. I find `""` to be more intuitive than `nullptr` but I do acknowledge this is personal preference. Using string shortens code in programs using these headers and importantly provides consistency between the arguments of the function. I would rather add a `string` overload for the remaining compile function (if this is possible) and deprecate the `const char*` one. I prefer it this way, but I do need the function, so if you insist I will change it to get this merged.

tests/test_openclhpp.cpp
```diff
@@ -2149,6 +2149,220 @@ void testBuildProgramSingleDevice(void)
...
+    cl_program         program,
+    cl_program_info    param_name,
+    size_t             param_value_size,
+    void *             param_value,
+    size_t *           param_value_size_ret,
+    int num_calls)
+{
+    (void) num_calls;
+
+    TEST_ASSERT_EQUAL(param_name, CL_PROGRAM_DEVICES);
```
bashbaug, line 2162: Fixes a CI error due to an unused parameter:

```suggestion
    TEST_ASSERT_EQUAL(program, make_program(0));
    TEST_ASSERT_EQUAL(param_name, CL_PROGRAM_DEVICES);
```

bashbaug:
Merging as discussed in the April 9th teleconference - thanks!

PR #287, opened 2024-02-16
Fix self-assign build error

lakshmih:


bashbaug (approved):
I'll merge this as soon as the CI builds complete, but we should really merge this PR ASAP so we don't keep encountering and fixing the same issues: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/251

It fixes all of the warnings and enables warnings-as-errors in CI so we will catch problems like these earlier.

PR #290, opened 2024-04-15
add string overloads for options and kernel names

### LINKED ISSUE #289: should we use strings or char pointers to pass string data?
See the comment on this PR: https://github.com/KhronosGroup/OpenCL-CLHPP/pull/286#issuecomment-2031694061

For some APIs that accept string arguments we currently pass the string data as a `char*`, see for example the `cl::Program::build` API:

https://github.com/KhronosGroup/OpenCL-CLHPP/blob/fc6123cb7c44b21f31f62823a372481e5f8cd72b/include/CL/opencl.hpp#L6640-L6644

Should we pass these as strings instead, or at least provide overloads that accept strings?

bashbaug:
fixes #289 

Adds string overloads for the program options and kernel name.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -1143,7 +1143,6 @@ inline cl_int getInfoHelper(Func f, cl_uint name, vector<vector<unsigned char>>*
         }
     }
 
-
     return CL_SUCCESS;
 }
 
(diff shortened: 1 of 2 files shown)


Snektron (approved):

bashbaug:
Merging as discussed in the April f2f.

PR #295, opened 2024-06-19
remove removed external memory and semaphore extensions

bashbaug:
We're removing the `cl_khr_external_memory_dx` and `cl_khr_external_semaphore_dx_fence` extensions due to lack of implementations and test coverage,so the handle types added by these extensions should also be removed from the C++ headers.

See: https://github.com/KhronosGroup/OpenCL-Docs/pull/1160

If we decide to keep the handle types for these extensions in the C++ headers for whatever reason, we should still keep the first commit, which adds the necessary include guards around each of the external memory handle types.

Fixes: https://github.com/KhronosGroup/OpenCL-Headers/issues/255


kpet (approved):
LGTM

EwanC (approved):
I'm not very familiar with this repo so can't review too deeply, but this change makes sense in the context of the other PRs that have merged.

bashbaug:
Looks like this is still failing, even after merging the latest CI/CD changes from the main branch.  The main branch is failing too, though.

I'm going to go ahead and merge this because it is causing build failures with the latest headers.  We still need to fix the CI issues, though...

PR #296, opened 2024-06-25
Add missing CL_HPP_ENABLE_EXCEPTIONS ifdeffery

gorazd-sumkovski-arm:


bashbaug (approved):
LGTM!

I suppose one could argue that `CL_HPP_ENABLE_EXCEPTIONS` should only affect the behavior for OpenCL errors, but I agree it is preferable not to throw exceptions explicitly when `CL_HPP_ENABLE_EXCEPTIONS` is not defined, also.

bashbaug:
Our CI has been having issues lately.  I don't think your change has caused these failures...

kpet (approved):

bashbaug (approved):

bashbaug:
This has been open for a while now, so I'm going to go ahead and merge this.  If there are any issues, we can always fix them after the fact.

PR #298, opened 2024-07-31
Update clUpdateMutableCommandsKHR binding

EwanC:
In OpenCL-Docs PR https://github.com/KhronosGroup/OpenCL-Docs/pull/1045 the API for `cl_khr_command_buffer_mutable_dispatch` API `clUpdateMutableCommandsKHR` changed in a breaking way.

When the headers update OpenCL-Headers PR https://github.com/KhronosGroup/OpenCL-Headers/pull/245 the bindings will break if they are not updated to reflect this change.

Bindings updated in this PR to pass to arrays of templated length to the C++ method. The underlying `.data()` pointers of these parameters can be passed through to the OpenCL API. I don't think any tests for this API currently exist in the repo to be updated.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -11549,6 +11549,8 @@ public:
     }
 
 #if defined(cl_khr_command_buffer_mutable_dispatch)
+#if CL_KHR_COMMAND_BUFFER_MUTABLE_DISPATCH_EXTENSION_VERSION <                 \
+    CL_MAKE_VERSION(0, 9, 2)
     cl_int updateMutableCommands(const cl_mutable_base_config_khr* mutable_config)
     {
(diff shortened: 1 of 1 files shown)


bashbaug (commented):
These changes look good, but they'll break without the corrosponding header updates.  Should we use the extension versioning mechanism to either a) only expose this function when the headers are sufficiently new, or b) select between both versions, for some time at least?

EwanC:
> These changes look good, but they'll break without the corrosponding header updates. Should we use the extension versioning mechanism to either a) only expose this function when the headers are sufficiently new, or b) select between both versions, for some time at least?

I forgot to reply when I made this change, but I updated to option b)

bashbaug (approved):
Merging as discussed in the September 3rd teleconference + email.

PR #304, opened 2024-08-01
Disable gcc-11 on macos

bashbaug:
There is a compiler issue with __has_cpp_attribute that is causing these CI configs to fail, so disable these configs (for now, at least).


Kerilk (approved):
LGTM.

bashbaug:
Merging to unblock CI.

PR #305, opened 2024-08-15
Add properties parameter to all command-buffer commands

EwanC:
Updates to compile with header change https://github.com/KhronosGroup/OpenCL-Headers/pull/260


bashbaug:
I think it would be better to keep the current functions as they currently are, with no "properties" argument, and then add overloads with a "properties" argument.  This has two main advantages:

1. Any code previously written for the older functions continues to work without any changes.
2. When the properties argument is not required - which will be the common case initially, at least - we won't need to pass as many arguments.

We could even skip adding the overloads with the "properties" argument for now, at least for the command buffer APIs where no properties are currently defined.

EwanC:
> I think it would be better to keep the current functions as they currently are, with no "properties" argument, and then add overloads with a "properties" argument. This has two main advantages:
> 
>     1. Any code previously written for the older functions continues to work without any changes.
> 
>     2. When the properties argument is not required - which will be the common case initially, at least - we won't need to pass as many arguments.
> 
> 
> We could even skip adding the overloads with the "properties" argument for now, at least for the command buffer APIs where no properties are currently defined.

Avoiding breaking user code if we can avoid it makes sense to me. Updated and left out the overloaded properties variant.

bashbaug (approved):
Thanks!  LGTM.

bashbaug:
Merging as discussed in the September 3rd teleconference + email.

The CI failures appear to be real, but are unrelated to these code changes.

PR #306, opened 2024-09-09
Fix updateMutableCommands template to use std::array::size() and const pointers

banan328:
This pull request fixes this commit https://github.com/KhronosGroup/OpenCL-CLHPP/pull/298:

- Replaced configs.length() with configs.size() to correctly retrieve the array size.
- Changed configs to std::array<const void*, ArrayLength> for better const-correctness.
- Passed config_types.data(), configs.data() directly to pfn_clUpdateMutableCommandsKHR.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -11876,14 +11876,14 @@ public:
     template <int ArrayLength>
     cl_int updateMutableCommands(std::array<cl_command_buffer_update_type_khr,
                                             ArrayLength> &config_types,
-                                 std::array<void *, ArrayLength> &configs) {
+                                 std::array<const void *, ArrayLength> &configs) {
         if (pfn_clUpdateMutableCommandsKHR == nullptr) {
             return detail::errHandler(CL_INVALID_OPERATION,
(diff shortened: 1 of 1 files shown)


bashbaug:
Could you add a test for this function?

(It would have been good to do that previously, since it probably would have caught these bugs earlier.)

EwanC:
> Could you add a test for this function?
> 
> (It would have been good to do that previously, since it probably would have caught these bugs earlier.)

There is an open PR https://github.com/KhronosGroup/OpenCL-CLHPP/pull/276 with some testing, will probably be a bit stale and need updated but could prioritize getting that in as a follow-up.

EwanC:
> > Could you add a test for this function?
> > (It would have been good to do that previously, since it probably would have caught these bugs earlier.)
> 
> There is an open PR #276 with some testing, will probably be a bit stale and need updated but could prioritize getting that in as a follow-up.

I've created https://github.com/KhronosGroup/OpenCL-CLHPP/pull/308 which updates and combines the work. So would supersede this PR and #276  if we merged it

bashbaug:
Fixed by #308 - closing.

PR #308, opened 2024-10-03
Combined mutable dispatch fixes

EwanC:
This PR combines the API fixes from https://github.com/KhronosGroup/OpenCL-CLHPP/pull/306 with the test additions from https://github.com/KhronosGroup/OpenCL-CLHPP/pull/276 such that there are tests that verify the updated API.

I've cherry picked the commits to correctly attribute the original authors, then added a merge commit and test fixup commit myself.


bashbaug:
The CI issue should be fixed now if you merge in the latest changes from main.

Is that the only problem that needs solving before marking this PR ready for review?

EwanC:
> The CI issue should be fixed now if you merge in the latest changes from main.
> 
> Is that the only problem that needs solving before marking this PR ready for review?

Yes this is ready for review, I've merged in `main` and taken the PR out of draft

bashbaug (approved):
Thanks, LGTM.

I wonder if we should eventually add an `updateMutableCommands` overload that takes `std::vectors` also (not in this PR, though).

Issue #293, opened 2024-05-14
New MSVC warning in `cl::copy`


MathiasMagnus:
Previously MSVC threw a warning for using naked pointers with algorithms in Debug builds without using a vendor utility promising that the pointers are valid over a certain range. Latest VS deprecated this feature and now it issues a deprecation warning.

```
D:\Develop\GPGPU1\build\msbuild-msvc\_deps\opencl-sdk-src\external\OpenCL-CLHPP\include\CL\opencl.hpp(10031,50): warning C4996: 'stdext::checked_array_iterator<DataType *>::checked_array_iterator': warning STL4043: stdext::checked_array_iterator, stdext::unchecked_array_iterator, and related factory functions are non-Standard extensions and will be removed in the future. std::span (since C++20) and gsl::span can be used instead. You can define _SILENCE_STDEXT_ARR_ITERS_DEPRECATION_WARNING or _SILENCE_ALL_MS_EXT_DEPRECATION_WARNINGS to suppress this warning. [D:\Develop\GPGPU1\build\msbuild-msvc\OpenCL\C++-API\SAXPY\OpenCL-Cpp-SAXPY.vcxproj]
```

Kerilk:
This is breaking the CI:
Here are the offending code lines:
https://github.com/KhronosGroup/OpenCL-CLHPP/blob/6db44b8db11952b53d271b6d1657ac5d04a45871/include/CL/opencl.hpp#L10302-L10310

bashbaug:
https://github.com/KhronosGroup/OpenCL-CLHPP/pull/309 might fix this issue.

I don't have easy access to all versions of visual studio to test, but this at least "fixes" our CI failures.

bashbaug:
#309 fixes this issue.  Closing.

PR #309, opened 2024-10-03
disable use of stdext::checked_array_iterator for VS2019 and newer

bashbaug:
Possible fix for https://github.com/KhronosGroup/OpenCL-CLHPP/issues/293.

It seems that the use of `stdext::checked_array_iterator` is not required for VS2019 and newer.  So, add `_MSC_VER` checks to fall back to the generic version for newer versions of Visual Studio.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -10299,15 +10299,15 @@ inline cl_int copy( const CommandQueue &queue, IteratorType startIterator, Itera
     if( error != CL_SUCCESS ) {
         return error;
     }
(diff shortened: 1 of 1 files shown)


MathiasMagnus (approved):
Checked back to VS 2015. 2015 (toolset v140) issues a false warning on /W4, but we can't fix that. (I argues about an unreferenced formal parameter to a function, but the parameter is totally used. No ifdefs, no control flow...) This change however seems fine. Toolsets v141, v142, v143 all pass warning free.

PR #313, opened 2024-10-25
Fix -Wcast-qual warnings

steffen-kiess:
Cast pointers to the correct type to avoid -Wcast-qual warnings

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -3795,7 +3795,7 @@ public:
 
         return detail::errHandler(
             ::clWaitForEvents(
-                (cl_uint) events.size(), (events.size() > 0) ? (cl_event*)&events.front() : nullptr),
+                (cl_uint) events.size(), (events.size() > 0) ? (const cl_event*)&events.front() : nullptr),
             __WAIT_FOR_EVENTS_ERR);
     }
 };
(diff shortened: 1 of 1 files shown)


include/CL/opencl.hpp
```diff
@@ -9392,9 +9392,9 @@ class CommandQueue : public detail::Wrapper<cl_command_queue>
                 object_, userFptr, args.first, args.second,
                 (mem_objects != nullptr) ? (cl_uint) mem_objects->size() : 0,
                 (mem_objects->size() > 0 ) ? reinterpret_cast<const cl_mem *>(mem_objects->data()) : nullptr,
-                (mem_locs != nullptr && mem_locs->size() > 0) ? (const void **) &mem_locs->front() : nullptr,
+                (mem_locs != nullptr && mem_locs->size() > 0) ? const_cast<const void**>(&mem_locs->front()) : nullptr,
```
bashbaug, line 9395: Needing to use `const_cast` here is unfortunate, but I don't think there is any way to avoid it.

Linking with https://github.com/KhronosGroup/OpenCL-Docs/issues/9 for completeness.

bashbaug (approved):

bashbaug:
Merging as discussed in the October 29th teleconference.

PR #314, opened 2024-10-25
Add support for CL_HPP_OPENCL_API_WRAPPER macro

steffen-kiess:
Add support for a macro `CL_HPP_OPENCL_API_WRAPPER` which will be used to wrap all OpenCL API calls.This can be used e.g. to load the OpenCL library dynamically with `dlopen()`/`dlsym()`.

This is similar to <https://github.com/KhronosGroup/OpenCL-CLHPP/pull/170>, but uses a macro which is called with the API function name instead of a namespace. This macro can e.g. call a function which will open the OpenCL library if needed, look up the function name and return a pointer to API function.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -206,6 +206,11 @@
  *
  *   Enable the cl_khr_il_program extension.
  *
+ * - CL_HPP_OPENCL_API_WRAPPER
+ *
+ *   A macro which is used to wrap all OpenCL API symbols. This can be
+ *   used e.g. to load the OpenCL library dynamically with dlopen()/dlsym().
+ *
  *
  * \section example Example
  *
@@ -391,6 +396,12 @@
  */
 #ifndef CL_HPP_
 #define CL_HPP_
+
+#ifdef CL_HPP_OPENCL_API_WRAPPER
+#define CL_(name) CL_HPP_OPENCL_API_WRAPPER(name)
+#else
+#define CL_(name) ::name
+#endif
 
 /* Handle deprecated preprocessor definitions. In each case, we only check for
  * the old name if the new name is not defined, so that user code can define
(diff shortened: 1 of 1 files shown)


include/CL/opencl.hpp
```diff
@@ -392,6 +397,14 @@
 #ifndef CL_HPP_
 #define CL_HPP_
 
+#ifdef CL_HPP_OPENCL_API_WRAPPER
+#define _CL(name) CL_HPP_OPENCL_API_WRAPPER(name)
+#define _CL_PTR(name) CL_HPP_OPENCL_API_WRAPPER(name)
+#else
+#define _CL(name) ::name
+#define _CL_PTR(name) &::name
```
bashbaug, line 405: Do we need both `_CL` and `_CL_PTR`?  I can't figure out why `_CL_PTR` is needed, and it seems like we could use `_CL` in all cases.

If we do need `_CL_PTR` for some reason, should there be a separate wrapper macro for it as well?  I'd prefer not to do this, and would much rather eliminate `_CL_PTR`, but we should stay symmetric regardless.

One file nit: identifiers beginning with an underscore and a capital letter are considered reserved IIRC, so although I appreciate the brevity we should use a different spelling than `_CL`.

steffen-kiess: For some reason I was thinking that a `&` is needed when taking a function pointer, but I think that's only the case for pointer-to-members, not for normal function pointers, so using `_CL` everywhere should work fine, yes.

Also, I'll rename `_CL` to `CL_` which is not reserved.

steffen-kiess: Ok, I've updated the PR.

bashbaug (approved):
Thank you for your contribution!

Merging as discussed in the October 29th teleconference.

PR #317, opened 2024-11-11
Fix more -Wcast-qual warnings

steffen-kiess:
Cast pointers to the correct type to avoid -Wcast-qual warnings. The previous fix (#313) was missing some casts in template methods, but clang issues warnings for there methods even if the templates are not instantiated.


bashbaug (approved):
Thank you for these fixes!

PR #321, opened 2025-03-26
Fix typo

ahesham-arm:
`CL_KHR_COMMAND_BUFFER_VERSION` is incorrect, should be `CL_KHR_COMMAND_BUFFER_EXTENSION_VERSION` instead, which is defined in `CL/cl_ext.h`. This fixes a compilation error seen with `-Wundef` and `-Werror`.

PR #322, opened 2025-03-27
Enable beta extensions

ahesham-arm:


bashbaug:
Thanks for the PR!  I started taking a look at this yesterday after I saw the CI was failing for #321, but I didn't get very far.

Instead of a cmake option, would it be better to enable beta extensions unconditionally for the tests?

https://github.com/KhronosGroup/OpenCL-CLHPP/blob/02cbac343bafbb402bc3c61924e51171c681d6bf/tests/CMakeLists.txt#L94-L95

My rationale is:

- I don't think the tests can possibly work without beta extensions due to the way the mock files are created.
- We don't have cmake options for anything else like this, such as `CL_TARGET_OPENCL_VERSION`.

I'd be OK adding something like a `CL_HPP_ENABLE_BETA_EXTENSIONS` in addition, if we choose to.  This would be similar to how we have `CL_HPP_TARGET_OPENCL_VERSION` to set `CL_TARGET_OPENCL_VERSION`.  Or, we could just have clients define `CL_ENABLE_BETA_EXTENSIONS` directly before including `opencl.hpp` if they want to enable beta extensions.

ahesham-arm:
> Thanks for the PR! I started taking a look at this yesterday after I saw the CI was failing for #321, but I didn't get very far.
> 
> Instead of a cmake option, would it be better to enable beta extensions unconditionally for the tests?

Well, looks like we both reached the same conclusion at the same time :)

bashbaug (approved):

bashbaug:
Test-only fix, not waiting to merge.

PR #325, opened 2025-06-05
initialize return value for deterministic queries

bashbaug:
This is a change I have been using locally for some time now that I've found useful, so I thought I'd see if it's something worth adding to the upstream C++ bindings.  In short, for queries that return the result of the query directly (vs. through a pointer), the value that is returned is uninitialized.  This means that if the query returns an error, the value that gets returned is most likely random and non-deterministic, which likely isn't what users expect.

With this change, the return values are always initialized, so even if an error occurs the query always returns a deterministic value.  The return value is not always meaningful, or identifiable as a default value, but it frequently is, and even when it isn't it's beneficial to return a consistent value than a random value.

### DIFF
--- a/include/CL/opencl.hpp
+++ b/include/CL/opencl.hpp
@@ -2588,7 +2588,7 @@ public:
     getInfo(cl_int* err = nullptr) const
     {
         typename detail::param_traits<
-            detail::cl_device_info, name>::param_type param;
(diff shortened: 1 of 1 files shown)


kpet (approved):

kpet:
Merging as discussed in 2025/09/19 teleconference.

PR #340, opened 2026-05-13
C++ binding updates for OpenCL 3.1

bashbaug:
Updates the C++ bindings for OpenCL 3.1.

This mostly involved adding support for the new clGetKernelSuggestedLocalWorkSize with an associated new test, plus a bit of tidy up.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "monthly"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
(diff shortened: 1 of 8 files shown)


Kerilk (approved):
Looking good to me. Took me a bit of time to get why `local` is initialized to `global` (you want the same dimentionality, so the copy constructor gives you that). I think I would prefer it to be initialized to zero, but at the right dimentionality, but that is a refactoring of the class that may be out of scope of this PR.

bashbaug:
> Took me a bit of time to get why `local` is initialized to `global` [...]

Yeah, I couldn't think of a way to initialize it to zero (or some other default value) without either expanding the NDRange class a bunch or needing a bunch of conditional code.  Initializing to `global` mostly does what I want it to do, so I went with it for now.

I've added a comment to avoid possible confusion in the future.  We can revisit if we ever decide to update the NDRange class.

